PR #75, opened 2024-10-15
TOK-317: fix CD for QA and testnet

jurajpiar:
## What

- adds an optional ENVIRONMENT env var to append to chain_id to get a correct config file
- removes options from the manual trigger and adds label-based trigger
- adds CD for testnet on rc tag
- verify each call to the node in deploy.sh

## Why

- to retrieve the correct config file
- to make it easy to deploy specific versions
- to automate testnet deployments

## Refs

- Task: [TOK-317](https://rsklabs.atlassian.net/browse/TOK-317)
- Follows [dApp CD](https://github.com/rsksmart/bim-dao-frontend/blob/develop/.github/workflows/qa.deploy.yaml)


antomor (approved):
LGTM

PR #79, opened 2024-10-22
chore: deploy contracts for qa dapp

antomor:
## What

- include configuration used for dApp QA deployment

## Why

- dApp QA env requires contracts configured differently


jmendiola222 (approved):
LGTM, but I've just asked Javi to take a look as we'll be more familiarized with this environment.

PR #84, opened 2024-10-25
Tok 333/contract permissions

DinizSa:
## What

- Encapsulates governance functionality by injecting the Governance smart contract directly as a dependency into contracts that need it. This centralizes governance logic and removes duplicated code.

Note: Implementing roles from OZ was tried, but because we only want 1 address per "role", and we don't expect to add many more roles, the extra gas cost was not justifiable. A simpler approach was chosen to hold the governance addresses.

## Why

- **Reduce Duplication**: Encapsulates governance logic, avoiding repetition across contracts.
- **Cleaner Dependencies**: Removes the need to pass `ChangeExecutor` around just to access the governor, simplifying dependencies.
- **Improves Permission testing**: Removed the need for `ChangeExecutorMock.isAuthorizedChanger { return true; }`, which obscured permission tests.

## Testing

- Unit tests have been updated & added, and all previous flows should work as before.

## Refs

- [Jira Ticket](https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?assignee=712020%3A42dabdd3-0b32-4060-be21-62cfd46a1751&selectedIssue=TOK-333)
- [Notion page with issues analysis](https://www.notion.so/iovlabs/TOK-333-contracts-permissions-123c132873f9808484f3f69c0f9e0542)


src/governance/Governed.sol
```diff
@@ -1,80 +1,205 @@
...
     // ---- External Functions -----
     // -----------------------------
 
     /**
-     * @notice maintains Governed interface. Returns governed address
+     * @notice Allows the governor to update its own role to a new address.
+     * @param governor_ The new governor address.
+     * @dev Reverts if caller is not the current governor.
+     */
+    function updateGovernor(address governor_) public onlyGovernor {
```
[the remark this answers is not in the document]
franciscotobar: Did we have this conversation?

src/governance/GovernanceManager.sol
```diff
@@ -0,0 +1,204 @@
...
+    // ---- External Functions -----
+    // -----------------------------
+
+    /**
+     * @notice Function to be called to make the changes in changeContract
+     * @dev reverts if is not called by the Governor
+     * @param changeContract_ Address of the contract that will execute the changes
+     */
+    function executeChange(IChangeContractRootstockCollective changeContract_) external onlyGovernor {
+        _updateChanger(address(changeContract_));
```
fruiz08, line 81: I would choose a different name to distinguish from the others(foundation, kycApprover) because this one is an atomic update.
maybe smt like:
address public authorizedChanger || validChanger;
_authorizeChanger()

fruiz08 (approved):
LGTM

nmagariang (approved):
LGTM

src/governance/GovernanceManager.sol
DinizSa: pushed 👍

antomor (approved):
LGTM

PR #90, opened 2024-10-29
refactor(whitelist): whitelist could be before activate

fruiz08:


src/SponsorsManager.sol
```diff
@@ -301,7 +301,7 @@ contract SponsorsManager is BuilderRegistry {
         internal
         returns (uint256 newSponsorTotalAllocation_, uint256 newTotalPotentialReward_)
     {
-        if (gaugeToBuilder[gauge_] == address(0)) revert GaugeDoesNotExist();
+        if (!isGaugeActivated(gauge_)) revert GaugeDoesNotExist();
```
franciscotobar, line 304: I guess that we will need to change the revert error.

src/BuilderRegistry.sol
```diff
@@ -378,6 +380,14 @@ abstract contract BuilderRegistry is EpochTimeKeeper, Ownable2StepUpgradeable {
         return _kickbackData.previous;
     }
 
+    /**
+     * @notice return true if gauge was created and activated
+     */
+    function isGaugeActivated(Gauge gauge_) public view returns (bool) {
```
franciscotobar, line 386:
```suggestion
    function isGaugeReady(Gauge gauge_) public view returns (bool) {
```
Or maybe another, more descriptive name.

src/BuilderRegistry.sol
```diff
@@ -53,11 +55,12 @@ abstract contract BuilderRegistry is EpochTimeKeeper, Ownable2StepUpgradeable {
     // ---------- Structs ----------
     // -----------------------------
     struct BuilderState {
+        bool activated;
```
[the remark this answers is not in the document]
fruiz08: yeap, it is to avoid foundation to revoke KYC and later use the activateBuilder allowing them to change the kickback

src/BuilderRegistry.sol
```diff
@@ -378,6 +380,14 @@ abstract contract BuilderRegistry is EpochTimeKeeper, Ownable2StepUpgradeable {
         return _kickbackData.previous;
     }
 
+    /**
+     * @notice return true if gauge was created and activated
+     */
+    function isGaugeActivated(Gauge gauge_) public view returns (bool) {
+        address _builder = gaugeToBuilder[gauge_];
+        return _builder != address(0) && builderState[_builder].activated;
```
franciscotobar, line 388: If we do this, is the cost affected?
```suggestion
        if (address(_gauge) == address(0)) revert BuilderDoesNotExist();
        if (!builderState[builder_].activated) revert NotActivated();
```

fruiz08: not too much, it is just one more if

src/SponsorsManager.sol
```diff
@@ -261,7 +261,7 @@ contract SponsorsManager is BuilderRegistry {
     function claimSponsorRewards(Gauge[] memory gauges_) external {
         uint256 _length = gauges_.length;
         for (uint256 i = 0; i < _length; i = UtilsLib._uncheckedInc(i)) {
-            if (gaugeToBuilder[gauges_[i]] == address(0)) revert GaugeDoesNotExist();
+            if (!isGaugeActivated(gauges_[i])) revert GaugeDoesNotExist();
```
nmagariang, line 264: If we're making the distinction between `NotExist` and `NotActivated ` (which I prefer) we should do that here too, right?

src/BuilderRegistry.sol
jmendiola222: That's the meaning Product has been using the whitelisig term for, so I'm ok with it.

jmendiola222: It can be address on other PR tho

jmendiola222 (approved):
LGTM

franciscotobar (approved):
LGTM.

@antomor I think that we just need to create a task to address the naming, but it can be done in the renaming task. Since now whitelisted is when you are `KYCApproved` and have the `CommunityAproval` which is what whitelisting meant before.

PR #92, opened 2024-10-31
fix: remove SponsorsManager as incentivizer in tests

nmagariang:
## What

- `SponsorsManager` is used as incentivizer in several tests, but approving `rewardToken`s for incentives overlaps with max reward approval in whitelisting process. Since `SponsorsManager` won't be incentivizing gauges directly in real life, it's safe to simply remove it from the tests without taking any more precautions


jmendiola222 (approved):
LGTM

PR #93, opened 2024-11-04
TOK-365: claim rewards by asset

franciscotobar:
## What

- Claim all rewards by asset

## Why

- There was no option to claim rewards by individual asset

## Testing

- Unit tests have been updated & added, and all previous flows should work as before.

## Refs

- [TOK-365](https://rsklabs.atlassian.net/browse/TOK-365)

### DIFF
--- a/src/SponsorsManager.sol
+++ b/src/SponsorsManager.sol
@@ -226,6 +226,20 @@ contract SponsorsManager is BuilderRegistry {
     }
 
     /**
+     * @notice claims sponsor rewards from a batch of gauges
+     * @param gauges_ array of gauges to claim
+     * @param rewardToken_ address of the token rewarded
+     *  address(uint160(uint256(keccak256("COINBASE_ADDRESS")))) is used for coinbase address
+     */
+    function claimSponsorRewards(address rewardToken_, Gauge[] memory gauges_) external {
+        uint256 _length = gauges_.length;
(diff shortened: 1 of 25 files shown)


src/SponsorsManager.sol
```diff
@@ -231,6 +231,20 @@ contract SponsorsManager is BuilderRegistry {
         }
     }
 
+    /**
+     * @notice claims sponsor rewards from a batch of gauges
+     * @param gauges_ array of gauges to claim
+     * @param rewardToken_ address of the token rewarded
+     *  address(uint160(uint256(keccak256("COINBASE_ADDRESS")))) is used for coinbase address
+     */
+    function claimSponsorRewards(address rewardToken_, Gauge[] memory gauges_) external {
```
franciscotobar, line 240: Im not sure if we want to add something like we did with the incentivize:

- [rewardToken](https://github.com/RootstockCollective/collective-rewards-sc/blob/main/src/gauge/Gauge.sol#L342)
- [coinbase](https://github.com/RootstockCollective/collective-rewards-sc/blob/main/src/gauge/Gauge.sol#L366)

nmagariang: I believe that separation in the gauge came from the fact that we could incentivize with the wrong token/send coinbase not accounted for, so besides being more explicit regarding the options to claim, I think it's not really necessary here

nmagariang (approved):
LGTM!

antomor (approved):
LGTM

PR #94, opened 2024-11-04
fix: positive allocation on halted gauge

franciscotobar:
## What

_Issue_
If a gauge is `halted` in epoch `x` and in epoch `x+1` is `resumed`, we need to take into account the following requirements. 

1. We need to use the `epochDuration` to avoid penalizing the backers that stayed in the gauge during the time that was `halted`
2. By using `epochDuration` we allowed the `builders` to increase their allocation at the end of the `epoch`, boosting their `totalPotentialReward` since the calculation will be done with the assumption that those allocations were there for the whole epoch. 

_Fix_
Do not allow `gauge` to increase their allocation while is `halted`.

### DIFF
--- a/src/gauge/Gauge.sol
+++ b/src/gauge/Gauge.sol
@@ -314,7 +314,7 @@ contract Gauge is ReentrancyGuardUpgradeable {
 
         // to avoid dealing with signed integers we add allocation if the new one is bigger than the previous one
         uint256 _previousAllocation = allocationOf[sponsor_];
-        if (allocation_ >= _previousAllocation) {
+        if (allocation_ > _previousAllocation) {
             allocationDeviation_ = allocation_ - _previousAllocation;
             rewardSharesDeviation_ = allocationDeviation_ * timeUntilNextEpoch_;
             totalAllocation += allocationDeviation_;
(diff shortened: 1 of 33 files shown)


src/gauge/Gauge.sol
```diff
@@ -314,7 +314,7 @@ contract Gauge is ReentrancyGuardUpgradeable {
 
         // to avoid dealing with signed integers we add allocation if the new one is bigger than the previous one
         uint256 _previousAllocation = allocationOf[sponsor_];
-        if (allocation_ >= _previousAllocation) {
+        if (allocation_ > _previousAllocation) {
```
franciscotobar, line 317: I had to change this for simplicity in the tests, I don't think that it would affect anything.

test/HaltedBuilderBehavior.t.sol
```diff
@@ -85,10 +85,10 @@ abstract contract HaltedBuilderBehavior is BaseTest {
 
     /**
      * SCENARIO: builder is halted in the middle of an epoch having allocation.
-     *  Alice modifies its allocation but the total reward shares don't change
+     *  Alice modifies reduce its allocation but the total reward shares don't change
```
jmendiola222, line 88:
```suggestion
     *  Alice reduces its allocation but the total reward shares don't change
```

jmendiola222 (approved):
LGTM

nmagariang (approved):
LGTM!

PR #95, opened 2024-11-05
[TOK-445] fix docgen workflow

jmendiola222:
## What

Modifies and fixes the docgen workflow, to only be run manually. 

## Why

PR review trigger was increasing the merge conflics and it doesn't adds much value.

## Testing

- manually run the workflow

## Refs

- https://rsklabs.atlassian.net/browse/TOK-445


antomor (approved):
LGTM

PR #98, opened 2024-11-07
TOK-453: Renaming in whitelisting flow

nmagariang:
## What

The whitelist process requires KYC approval and community approval. Currently the `whitelistBuilder` function handles the community approval, so this task is to rename that and everything related.

## Testing

- The flow itself didn't change, just the naming so existing tests still work

## Refs

- Task: [TOK-453](https://rsklabs.atlassian.net/browse/TOK-453)

### DIFF
--- a/src/BackersManagerRootstockCollective.sol
+++ b/src/BackersManagerRootstockCollective.sol
@@ -251,7 +251,7 @@ contract BackersManagerRootstockCollective is
         uint256 _length = gauges_.length;
         for (uint256 i = 0; i < _length; i = UtilsLib._uncheckedInc(i)) {
             // reverts if builder was not activated or approved by the community
-            _validateGauge(gauges_[i]);
+            _validateWhitelisted(gauges_[i]);
 
             gauges_[i].claimBackerReward(msg.sender);
         }
@@ -307,7 +307,7 @@ contract BackersManagerRootstockCollective is
         returns (uint256 newbackerTotalAllocation_, uint256 newTotalPotentialReward_)
     {
         // reverts if builder was not activated or approved by the community
-        _validateGauge(gauge_);
+        _validateWhitelisted(gauge_);
 
         (uint256 _allocationDeviation, uint256 _rewardSharesDeviation, bool _isNegative) =
             gauge_.allocate(msg.sender, allocation_, timeUntilNextCycle_);
@@ -439,7 +439,7 @@ contract BackersManagerRootstockCollective is
 
     /**
      * @notice approves rewardTokens to a given gauge
-     * @dev give full allowance when it is whitelisted and remove it when it is dewhitelisted
+     * @dev give full allowance when it is community approved and remove it when it is dewhitelisted
      * @param gauge_ gauge contract to approve rewardTokens
      * @param value_ amount of rewardTokens to approve
      */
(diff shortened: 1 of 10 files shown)


src/BackersManagerRootstockCollective.sol
```diff
@@ -249,7 +249,7 @@ contract BackersManagerRootstockCollective is
         uint256 _length = gauges_.length;
         for (uint256 i = 0; i < _length; i = UtilsLib._uncheckedInc(i)) {
             // reverts if builder was not activated or approved by the community
-            _validateGauge(gauges_[i]);
+            _isWhitelisted(gauges_[i]);
```
nmagariang, line 252: maybe `validateWhitelisting` or something similar?

DinizSa: `isWhitelisted` seems that the functions returns a boolean.
I kinda like `_validateGauge`, as it does not use the whitelisted terminology that we are moving away from, and is in checking if that gauge is valid.

src/BuilderRegistryRootstockCollective.sol
```diff
@@ -221,7 +221,7 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
         builderState[builder_].kycApproved = false;
 
         GaugeRootstockCollective _gauge = builderToGauge[builder_];
-        // if builder is whitelisted, it has a gauge associated
+        // if builder is communityApproved, it has a gauge associated
```
DinizSa, line 224: maybe `community approved` instead, for readability

src/BackersManagerRootstockCollective.sol
nmagariang: >because is prefix is more common when the function returns a boolean.

Exactly, that's why I was making that comment, we discussed before about leaving it with the `isWhitelisted` but it doesn't look right

>what does it mean for a builder to be whitelisted? is it kyc approved & community approved & activated? in other words, a fully valid builder?

That's right, being whitelisted means you have both approvals, that's why I want to leave at least here the `whitelisted` terminology

>validateWhitelisted or checkWhitelisted

Thanks, I'll go with one of them 🙌

DinizSa: great thanks

DinizSa:
LGTM

src/BuilderRegistryRootstockCollective.sol
```diff
@@ -494,7 +498,7 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
     /**
      * @notice reverts if builder was not activated or approved by the community
      */
-    function _validateGauge(GaugeRootstockCollective gauge_) internal view {
+    function _validateWhitelisted(GaugeRootstockCollective gauge_) internal view {
```
antomor, line 501: NIT: just a doubt about this, because it's not the Gauge that got whitelisted, but the Builder. Neverthless, I couldn't fund a different name to suggest 😓

antomor (approved):
LGTM

src/BuilderRegistryRootstockCollective.sol
antomor: yes, indeed and `validateGauge` doesn't make sense. Even `validateGaugeIsWhitelisted` doesn't make sense, so I think we should live with this 😅

jmendiola222 (approved):
LGTM

PR #99, opened 2024-11-07
Tok 353/migration to v2

DinizSa:
## What

- Allow builders migration from MVP to v2

## Testing

- Unit testing

## Refs

- [Jira Ticket](https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?assignee=712020%3A42dabdd3-0b32-4060-be21-62cfd46a1751&selectedIssue=TOK-353)

### DIFF
--- a/src/BuilderRegistryRootstockCollective.sol
+++ b/src/BuilderRegistryRootstockCollective.sol
@@ -166,26 +166,7 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
         external
         onlyKycApprover
     {
-        if (builderState[builder_].activated) revert AlreadyActivated();
-        builderState[builder_].activated = true;
-        builderState[builder_].kycApproved = true;
-        builderRewardReceiver[builder_] = rewardReceiver_;
-        // TODO: should we have a minimal amount?
-        if (rewardPercentage_ > _MAX_REWARD_PERCENTAGE) {
-            revert InvalidBuilderRewardPercentage();
-        }
-
-        // read from storage
-        RewardPercentageData memory _rewardPercentageData = builderRewardPercentage[builder_];
-
-        _rewardPercentageData.previous = rewardPercentage_;
-        _rewardPercentageData.next = rewardPercentage_;
-        _rewardPercentageData.cooldownEndTime = uint128(block.timestamp);
-
-        // write to storage
-        builderRewardPercentage[builder_] = _rewardPercentageData;
-
-        emit BuilderActivated(builder_, rewardReceiver_, rewardPercentage_);
+        _activateBuilder(builder_, rewardReceiver_, rewardPercentage_);
     }
 
     /**
@@ -239,15 +220,7 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
      * @return gauge_ gauge contract
      */
     function whitelistBuilder(address builder_) external onlyValidChanger returns (GaugeRootstockCollective gauge_) {
-        if (builderState[builder_].whitelisted) revert AlreadyWhitelisted();
-        if (address(builderToGauge[builder_]) != address(0)) revert BuilderAlreadyExists();
-
-        builderState[builder_].whitelisted = true;
-        gauge_ = _createGauge(builder_);
-
-        _rewardTokenApprove(address(gauge_), type(uint256).max);
-
-        emit Whitelisted(builder_);
+        return _whitelistBuilder(builder_);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/BuilderRegistryRootstockCollective.sol
```diff
@@ -537,6 +536,50 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
...
+    function _whitelistBuilder(address builder_) private returns (GaugeRootstockCollective gauge_) {
+        if (builderState[builder_].whitelisted) revert AlreadyWhitelisted();
+        if (address(builderToGauge[builder_]) != address(0)) revert BuilderAlreadyExists();
+
+        builderState[builder_].whitelisted = true;
+        gauge_ = _createGauge(builder_);
+
+        _rewardTokenApprove(address(gauge_), type(uint256).max);
+
+        emit Whitelisted(builder_);
```
jurajpiar, line 580: should we moved this to the public function so that it's not emitted upon migration which emits it's own event?

src/BuilderRegistryRootstockCollective.sol
```diff
@@ -537,6 +536,50 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
...
+        RewardPercentageData memory _rewardPercentageData = builderRewardPercentage[builder_];
+
+        _rewardPercentageData.previous = rewardPercentage_;
+        _rewardPercentageData.next = rewardPercentage_;
+        _rewardPercentageData.cooldownEndTime = uint128(block.timestamp);
+
+        // write to storage
+        builderRewardPercentage[builder_] = _rewardPercentageData;
+
+        emit BuilderActivated(builder_, rewardReceiver_, rewardPercentage_);
```
jurajpiar, line 564: same here (?)

src/BuilderRegistryRootstockCollective.sol
DinizSa: Is a good point, honestly I'm divided between both approaches, because a user is still being whitelisted and activated in v2, just through a different flow (migration instead of the usual flow). 
Also if a dapp is listening to the events `Whitelisted` or `BuilderActivated`, they would not catch the migrated users into the their system, which might be not ideal. wdyt?

src/BuilderRegistryRootstockCollective.sol
```diff
@@ -391,6 +369,27 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
         builderRewardPercentage[msg.sender] = _rewardPercentageData;
     }
 
+    /**
+     * @notice migrate mvp builder to the new builder registry
```
jmendiola222, line 373: I think V1 is used, `mvp` is just how we call it internally.

src/BuilderRegistryRootstockCollective.sol
```diff
@@ -391,6 +369,27 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
...
+     * @dev reverts if builder is not whitelisted in the simplifiedRewardDistributor
```
jmendiola222, line 374: Where is this reverting happening, I missed it.

src/BuilderRegistryRootstockCollective.sol
```diff
@@ -391,6 +369,27 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
...
+     * @param builder_ address of the builder
```
jmendiola222, line 375:
```suggestion
     * @param builder_ address of the builder whitelisted on the V1's SimplifiedRewardDistributor contract
```

src/BuilderRegistryRootstockCollective.sol
```diff
@@ -391,6 +369,27 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
...
+     * @param rewardAddress_ address of the builder reward receiver
```
jmendiola222, line 376:
```suggestion
     * @param rewardAddress_ address of the builder reward receiver whitelisted on the V1's SimplifiedRewardDistributor contract
```

src/BuilderRegistryRootstockCollective.sol
jmendiola222: I would leave them. For indexing purposes, you can always filter by this event and see all Builder history. The migration exclusion can be made by joining the two subsets.

src/BuilderRegistryRootstockCollective.sol
```diff
@@ -391,6 +369,27 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
...
+     * @param rewardPercentage_ reward percentage(100% == 1 ether)
+     */
+    function migrateBuilder(
+        address builder_,
+        address rewardAddress_,
+        uint64 rewardPercentage_
+    )
+        public
+        onlyFoundationTreasury
```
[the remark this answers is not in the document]
DinizSa: It makes sense, will do

src/BuilderRegistryRootstockCollective.sol
DinizSa: fixed 👍  I removed the v1 whitelisted check after the slack discussions, but forgot to remove the comment.

antomor (approved):
LGTM

nmagariang (approved):
LGTM!

jmendiola222 (approved):
LGTM

PR #100, opened 2024-11-08
fix: include params when deploying GovernanceManager from main script

antomor:
## What

- Include env vars when deploying GovernanceManager from the main script

## Why

- To follow the same approach used for the others


jmendiola222 (approved):
LGTM

PR #103, opened 2024-11-08
fix: exclude contracts from target senders

nmagariang:
## What

- There is some flakiness in the invariant testing that seems to be related to contracts being dealt with balance while acting as senders and thus affecting calculations

## Testing

- Invariant tests should pass


jmendiola222 (approved):
LGTM

PR #105, opened 2024-11-11
Set the tokenAmountPerEpoch

bezerrablockchain:
## What

Created state variables and corresponding set method to persist a default values to be used when Rewards needs to be sent with default values

## Testing

- Two basic tests using the new 03 methods were added

## Refs

- https://rsklabs.atlassian.net/browse/TOK-459


src/RewardDistributorRootstockCollective.sol
```diff
@@ -81,7 +86,7 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective
      * @param amountERC20_ amount of ERC20 reward token to send
      * @param amountCoinbase_ amount of Coinbase reward token to send
      */
-    function sendRewards(uint256 amountERC20_, uint256 amountCoinbase_) external payable onlyFoundationTreasury {
+    function sendRewards(uint256 amountERC20_, uint256 amountCoinbase_) public payable onlyFoundationTreasury {
```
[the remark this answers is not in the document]
bezerrablockchain: Coz' in the task specification we should call the `external sendRewards`, but in order to be called internally from another function it should be changed to `public`.
Its been called by new function `sendRewardsWithDefaultAmount`.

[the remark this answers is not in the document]
jmendiola222: But you could have used the internal fnc instead, right? `_sendRewards(...);` And so, avoiding the `onlyFoundationTreasury` verification twice.

test/RewardDistributorRootstockCollective.t.sol
```diff
@@ -125,4 +125,56 @@ contract RewardDistributorRootstockCollectiveTest is BaseTest {
         // THEN coinbase balance of gauge is 3 ether
         assertEq(address(gauge).balance, 3 ether);
     }
+
+    /**
+     * SCENARIO: sends rewards and starts the distribution with default amounts
```
jmendiola222, line 130: Do you think the "out of funds" failing scenario adds value? Meaning that the the `sendRewardsWithDefaultAmount();` is call a couple of times until the funds run out and you get an error.

src/RewardDistributorRootstockCollective.sol
```diff
@@ -97,14 +102,52 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective
...
+    // ---- External Functions -----
+    // -----------------------------
+
+    /**
+     * @notice sets the default reward amounts
+     * @dev reverts if is not called by foundation treasury address
+     * @param tokenAmount_ default amount of ERC20 reward token to send
+     * @param coinbaseAmount_ default amount of Coinbase reward token to send
+     */
+    function setDefaultRewardAmount(
```
jmendiola222, line 123: We are not emitting any event? Maybe we should? WDYT @antomor ?

src/RewardDistributorRootstockCollective.sol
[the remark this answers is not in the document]
bezerrablockchain: changed to use internal functions

test/RewardDistributorRootstockCollective.t.sol
bezerrablockchain: it could be.  But the task specification says to assume that the contract has enough funds

src/RewardDistributorRootstockCollective.sol
```diff
@@ -40,6 +40,11 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective
     /// @notice tracks amount of coinbase distributed per cycle
     mapping(uint256 cycleTimestampStart => uint256 amount) public rewardCoinbaseAmountPerCycle;
```
DinizSa, line 41: Can these unused state variables be deleted?

jmendiola222: Yes, missed that, right. They need to be removed. ;)

test/RewardDistributorRootstockCollective.t.sol
jmendiola222: I don't think having such an assumption exempt us from covering edge cases with tests. Maybe it was covered on the other already supported scenario, but we should be sure that in that case, it fails gracefully, reverting for example with an appropriate reason for the user to understand. It shouldn't fail silently for example and not completing either the coinbase nor the token transfer.

src/RewardDistributorRootstockCollective.sol
bezerrablockchain: Removed!

jmendiola222 (approved):
LGTM

PR #106, opened 2024-11-12
chore: change deep test job name

antomor:
## What

- change the deep test job name

## Why

- so we can remove it from the required checks (at the moment we have 2 tests jobs, so we either enable or disable both)


nmagariang (approved):

jmendiola222 (approved):
LGTM

PR #107, opened 2024-11-12
Tok 460/allow foundation to set the distribution window

DinizSa:
## What

- Allow the Foundation (or a valid changer) to set the distribution window

## Refs

- [Jira ticket](https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?assignee=712020%3A42dabdd3-0b32-4060-be21-62cfd46a1751&selectedIssue=TOK-460)


src/CycleTimeKeeperRootstockCollective.sol
```diff
@@ -112,6 +138,40 @@ abstract contract CycleTimeKeeperRootstockCollective is UpgradeableRootstockColl
...
+     * @dev only callable by the foundation
+     * @param newDistributionDuration_ new distribution window duration
+     */
+    function setDistributionDuration(uint32 newDistributionDuration_) external onlyFoundation {
+        // revert if the new distribution duration is too short
+        if (newDistributionDuration_ == 0) revert DistributionDurationTooShort();
+
+        // revert if the distribution duration is modified during the distribution window
+        uint256 _cycleStart = cycleStart(block.timestamp);
+        if (block.timestamp > _cycleStart && block.timestamp < _cycleStart + distributionDuration) {
```
jmendiola222, line 155: Shouldn't we also check for the new value?
`block.timestamp < _cycleStart + newDistributionDuration_`

test/SetDistributionDuration.t.sol
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * SCENARIO: Reverts if distribution duration is modified during an active distribution window
+     */
+    function test_RevertWhenModifiedDuringDistributionWindow() public {
+        // GIVEN the caller is authorized and an active distribution window is ongoing
+        vm.startPrank(foundation);
+
+        // Start a distribution cycle to put the contract within an active distribution window
```
jmendiola222, line 49: I don't undestand this comment not the need and correlation with the code following it.

src/CycleTimeKeeperRootstockCollective.sol
```diff
@@ -6,19 +6,34 @@ import { UtilsLib } from "./libraries/UtilsLib.sol";
 import { IGovernanceManagerRootstockCollective } from "./interfaces/IGovernanceManagerRootstockCollective.sol";
 
 abstract contract CycleTimeKeeperRootstockCollective is UpgradeableRootstockCollective {
-    uint256 internal constant _DISTRIBUTION_WINDOW = 1 hours;
+    error NotValidChangerOrFoundation();
+
+    modifier onlyValidChangerOrFoundation() {
+        if (!governanceManager.isAuthorizedChanger(msg.sender) && msg.sender != governanceManager.foundationTreasury())
```
jmendiola222, line 12: Why excluding the governor and not using `validateChanger`?

DinizSa: `governanceManager.isAuthorizedChanger` checks for both the governor and changer.
Calling`validateChanger` would revert, and we need to check for equality with other addresses.

src/CycleTimeKeeperRootstockCollective.sol
DinizSa: good catch, just pushed it

test/SetDistributionDuration.t.sol
DinizSa: That comment is about setting the distribution period and time ward into the middle of it.
I ended up removing that section of the code: I was setting the distribution duration in the test, although it was already set on contract construction. This was to not rely on the contract construction parameter. But to simplify the test, I just rely now on the value passed during construction.

src/CycleTimeKeeperRootstockCollective.sol
jmendiola222: Then I don't understand this test where you're pranking the governor and expecting it to fail: https://github.com/RootstockCollective/collective-rewards-sc/pull/107/files#diff-c0aed7c6761bd3039b7276487045f2521369fd3b96ff816c3c75a65e8914113eR112

nmagariang (approved):
LGTM!

PR #109, opened 2024-11-13
Remove legacy unused field

jmendiola222:
## What

removes unused field

## Why

it was depracated when GovernorManagar was introduced

## Testing

- Everything should keep working as it is before


nmagariang (approved):
LGTM!

PR #110, opened 2024-11-13
chore: add tests to move allocations with allocateBatch

nmagariang:
## What

- Add tests to verify that allocations can be moved from one builder to the other when all votes are in use

## Why

- Backers might want to move votes from one builder to the other in a single transaction

### DIFF
--- a/test/BackersManagerRootstockCollective.t.sol
+++ b/test/BackersManagerRootstockCollective.t.sol
@@ -272,6 +272,99 @@ contract BackersManagerRootstockCollectiveTest is BaseTest {
     }
 
     /**
+     * SCENARIO: alice has all votes staked and wants to move some from one gauge to another using allocateBatch
+     */
+    function test_ReallocateVotes() public {
+        // GIVEN alice has 100_000 ether in stakingToken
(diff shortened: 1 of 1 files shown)


test/BackersManagerRootstockCollective.t.sol
```diff
@@ -271,6 +271,96 @@ contract BackersManagerRootstockCollectiveTest is BaseTest {
...
+        assertEq(gauge.allocationOf(alice), 0);
+        assertEq(gauge2.allocationOf(alice), 100_000 ether);
+    }
+
+    /**
+     * SCENARIO: alice has all votes staked and wants to move votes from a revoked gauge to another one using
+     * allocateBatch
+     */
+    function test_ReallocateVotesFromRevokedBuilder() public {
+        // GIVEN alice has 100_000 ether in stakingToken
```
jmendiola222, line 315: I think the test should assert this as well, given this is not part of this test in particular, if someone increases that value in the baseTest, this test will give false positives silently.

jmendiola222 (approved):
Beside minor sugestion, looks good

PR #111, opened 2024-11-13
[TOK-454] feat: add veiw for estimated remaining rewards in cycle

nmagariang:
## What

- There is a requirement to show estimated rewards left to earn for backers withing a cycle

## Testing

-  New tests added

## Refs

- [Task](https://rsklabs.atlassian.net/browse/TOK-454)

### DIFF
--- a/src/gauge/GaugeRootstockCollective.sol
+++ b/src/gauge/GaugeRootstockCollective.sol
@@ -157,6 +157,21 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
     }
 
     /**
+     * @notice gets the estimated amount of rewardToken left to earn for a backer in current cycle
+     * @param rewardToken_ address of the token rewarded
+     *  address(uint160(uint256(keccak256("COINBASE_ADDRESS")))) is used for coinbase address
+     * @param backer_ address of the backer
+     */
+    function estimatedBackerRewards(address rewardToken_, address backer_) external view returns (uint256) {
+        // No allocations or a new cycle started without a distribution
+        if (totalAllocation == 0 || backersManager.periodFinish() <= block.timestamp) {
+            return 0;
+        }
+        // [N] = [N] * [N] / [N]
+        return _left(rewardToken_) * allocationOf[backer_] / totalAllocation;
+    }
+
+    /**
      * @notice gets amount of rewardToken earned for a backer
      * @param rewardToken_ address of the token rewarded
      *  address(uint160(uint256(keccak256("COINBASE_ADDRESS")))) is used for coinbase address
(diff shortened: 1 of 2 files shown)


src/gauge/GaugeRootstockCollective.sol
```diff
@@ -156,6 +156,21 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
         return rewardData[rewardToken_].backerRewardPerTokenPaid[backer_];
     }
 
+    /**
+     * @notice gets the estimated amount of rewardToken left to earn for a backer in current cycle
+     * @param rewardToken_ address of the token rewarded
+     *  address(uint160(uint256(keccak256("COINBASE_ADDRESS")))) is used for coinbase address
+     * @param backer_ address of the backer
+     */
+    function estimatedBackerRewards(address rewardToken_, address backer_) public view returns (uint256) {
```
jmendiola222, line 165: are we expecting to use this view internally? If not, it might be better just external?

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -190,9 +205,7 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
      *  address(uint160(uint256(keccak256("COINBASE_ADDRESS")))) is used for coinbase address
      */
     function left(address rewardToken_) external view returns (uint256) {
-        RewardData storage _rewardData = rewardData[rewardToken_];
-        // [N] = ([N] - [N]) * [PREC] / [PREC]
-        return UtilsLib._mulPrec(backersManager.periodFinish() - block.timestamp, _rewardData.rewardRate);
+        return _left(rewardToken_);
```
jmendiola222, line 208: why do we need a internal fnc? Shouldn't just this public and that's it?

src/gauge/GaugeRootstockCollective.sol
nmagariang: I don't think so, I'll change it to external 👍

src/gauge/GaugeRootstockCollective.sol
nmagariang: Yeah this change derived from some talks about doing it how we do in other functions, but I don't have a strong opinion about it, public without internal works for me too

jmendiola222: Can u point out another view where this is the case? Because for state changing functions it makes sense, as we usually apply modifiers for example on some of the external methods

jmendiola222 (approved):
Besides minor comments, lGTM

src/gauge/GaugeRootstockCollective.sol
franciscotobar: https://github.com/RootstockCollective/collective-rewards-sc/blob/45b93cd736e6b5f0f091e9277af313ae6330f8a0/src/gauge/GaugeRootstockCollective.sol#L173

This one for example:)

I was the one that thought that was the standard, but I don't have any preference:)

franciscotobar (approved):
LGTM

PR #114, opened 2024-11-15
[TOK-482] require min dust amount on incentivize

jmendiola222:
## What

Introduces a min min amount required on both coin base and tokens incentivize

## Why

calling this method zero, did not revert, it spams the protocol with useless tx and if done high frecuency, it might introduce rounding error. Demanding at least the 100 wei, the caller covers any potential dust error introduced. 
The 100 value is hard-coded, because we wanted to keep the gas impact of this check as minimal as possible. And as we said, it's only intended to cover for rounding errors, dust.

## Testing

Send any amount < 100 should fail.

## Refs

- [Jira ticket](https://rsklabs.atlassian.net/browse/TOK-482)


src/gauge/GaugeRootstockCollective.sol
```diff
@@ -40,6 +41,15 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
         _;
     }
 
+    /**
+     * @notice Prevent spamming incentives machanims with low values and introduce error
+     * @dev 100 shold at least cover the any potencial rounding error
```
nmagariang, line 46:
```suggestion
     * @notice prevents spamming of incentives mechanism with low values to introduce errors
     * @dev 100 should cover any potential rounding errors
```

nmagariang (approved):
LGMT!

antomor (approved):
LGTM

PR #121, opened 2024-11-27
[TOK-492] fix: collective rewards not beinig initialized on deploy

jmendiola222:
## What

Adds the previous missing `broadcast()` to actually sign and bradcast the Tx to the network

## Why

Collective rewards not being initialized on deployment with function `initializeCollectiveRewardsAddresses`, the call was there, but the Tx wasn't.

## Testing

Deploy on testnet and verify that now the contract is initialized with backers address.

## Refs

https://rsklabs.atlassian.net/browse/TOK-492
https://book.getfoundry.sh/cheatcodes/broadcast


antomor (approved):
@jmendiola222 Good catch, thank you! LGTM

PR #122, opened 2024-11-27
Relax testing solhint rules

jmendiola222:
## What

Turns off two lint rules that are not that important on testing context:
- "no-empty-blocks": "off",
-  "one-contract-per-file": "

And fixes/addresses same other trivial warning.

## Why

To clear the shell output from unnecessary warnings.

## Testing

`bun run lint`


nmagariang (approved):
LGTM!

PR #123, opened 2024-11-27
Add dust tolerance on invariants

jmendiola222:
## What

Introduce dust tolerance to gauge balance on invariant tests. 

## Why

As the deep run might include several cycle distributions, there might be duut left from previous ones.  

## Testing

let it :fire: those :computer: :computer:


nmagariang (approved):
LGTM!

PR #126, opened 2024-12-02
TOK-180: add network switch guide

jurajpiar:
## What

- adds a description of how to change networks 

## Why

- users may want to deploy to different networks

## Testing

- Follow the changes in the added readme lines to load the different chain configurations

## Refs

- ticket [TOK-108](https://rsklabs.atlassian.net/browse/TOK-180)


README.md
```diff
@@ -59,7 +59,7 @@ cd collective-rewards-sc
 bun install # install Solhint, Prettier, Hardhat and other Node.js deps
 ```
 
-> [!WARNING] `.env.<chain_id>` could be public, don't put your private key or mnemonic there. Use `.env` for that.
+> [!WARNING] > `.env.<chain_id>` could be public, don't put your private key or mnemonic there. Use `.env` for that.
```
jurajpiar, line 62: I'm not sure what to do about this.
1. we have a md linter, I'm not sure why we use prettier for this
2. not sure why we force prettier to line wrap since it [disabled by default as md renderers depend on specific line-breaks](https://prettier.io/docs/en/options#prose-wrap)
3. if we don't want to allow github rendering the notice blocks we should not use them

jmendiola222 (approved):
LGTM

README.md
franciscotobar: Then I think that we should avoid using prettier in the md documents.

PR #131, opened 2024-12-10
Code reorganizations

jmendiola222:
## What

- Makes BackerManager inherit from Interface
- Moves all BackerManager related contract files to `backerManager` folder

## Why

- makes de code more consistence and clear

## Testing

- all test should pass

## Refs

- This PR should be preferable merged after doc one


src/backersManager/CycleTimeKeeperRootstockCollective.sol
```diff
@@ -1,12 +1,16 @@
...
-import { IGovernanceManagerRootstockCollective } from "./interfaces/IGovernanceManagerRootstockCollective.sol";
+import { UpgradeableRootstockCollective } from "../governance/UpgradeableRootstockCollective.sol";
+import { UtilsLib } from "../libraries/UtilsLib.sol";
+import { IBackersManagerRootstockCollective } from "../interfaces/IBackersManagerRootstockCollective.sol";
+import { IGovernanceManagerRootstockCollective } from "../interfaces/IGovernanceManagerRootstockCollective.sol";
 import { Math } from "@openzeppelin/contracts/utils/math/Math.sol";
 
-abstract contract CycleTimeKeeperRootstockCollective is UpgradeableRootstockCollective {
+abstract contract CycleTimeKeeperRootstockCollective is
+    IBackersManagerRootstockCollective,
```
antomor, line 11: @jmendiola222 Why did you extend the interface here rather than in the BackersManager implementation directly?

jmendiola222: Because the interface include methods from all the hierarchy levels, from CycleTimeKeeper and BuilderRegistry. That's why I also thought it was better to group them under a "BackerManager" folder, to give some clarity that those two are not independent reusable abstractions, but more like BackerManager components broken down just to keep some responsibility separation.

jmendiola222:
Closing in favour of #135

PR #132, opened 2024-12-17
Nft support poc

jmendiola222:
## What

- This POC, explores the idea of using the NFT as backers, so that Gauge accounting can be 100% reused and only entry/exit points needs to be adapted.

## Why

- (the reason these changes are wanted)

## Testing

- (describe how to test the changes included)

## Refs

- (optional: include links to other issues, PRs, tickets, etc)


DinizSa:
following an offchain approach for now, we might get back to this approach later.

PR #130, opened 2024-12-05
TOK-516: document the contracts

antomor:
## What

- document the contracts included in the repo
- include a specification file to describe how the rewards are calculated

PR #133, opened 2024-12-18
TOK-265: Add PERMISSIONS.MD file

bezerrablockchain:
## What

- Added a Permissions file with info about permissions defined for the `GovernanceManagerRootstockCollective` contract

## Why

- For getting a more clear vision about what need to set or use to call this contract functions

## Refs

- https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?selectedIssue=TOK-265


docs/src/src/PERMISSIONS.md
antomor, line 1: I'd move this file into a spec folder, as in #130.

docs/src/src/PERMISSIONS.md
```diff
@@ -0,0 +1,91 @@
+# GovernanceManagerRootstockCollective Access Control
+This document describes the permissions, roles, and functionalities within the `GovernanceManagerRootstockCollective` contract.
+
+## User Roles and Abilities
+### Governor
+- Who: Defined by the `governor` state variable.
+- Abilities:
+-- Execute contracts implementing the `IChangeContractRootstockCollective` interface through executeChange.
+-- View validation of their role using `validateGovernor`.
```
antomor, line 9: I think that `--` is not rendered as expected, see [here](https://github.com/RootstockCollective/collective-rewards-sc/blob/tok-265/permissions-doc-gov/docs/src/src/PERMISSIONS.md)

docs/src/src/PERMISSIONS.md
```diff
@@ -0,0 +1,91 @@
...
+-- Update the `governor` address using `updateGovernor`.
+-- Update the `foundationTreasury` address using `updateFoundationTreasury`.
+-- Update the `kycApprover` address using `updateKYCApprover`.
+-- Validate their role with `validateAuthorizedChanger`.
+### Authorized Upgrader
+- Who: Address specified as `_authorizedChanger`, `governor`, or `upgrader`.
+- Abilities:
+-- Upgrade the contract implementation using `_authorizeUpgrade`.
+-- Validate their role using `validateAuthorizedUpgrader`.
+### KYC Approver
```
antomor, line 22: I'd try to be more specific here listing all the operations that the KYC approver can execute in the `BuilderRegistry` for instance

docs/src/src/PERMISSIONS.md
```diff
@@ -0,0 +1,91 @@
...
+- Who: Defined by the `kycApprover` state variable.
+- Abilities:
+-- Validate the KYC approver role using `validateKycApprover`.
+### Foundation Treasury
```
antomor, line 26: Same here, if you check where the `validateFoundationTreasury` is used, you will have a list of operations

antomor (changes requested):
@bezerrablockchain thank you for the changes. I'd try to be more specific about what each role can do, referring to the related functions of the other contracts. As a general comment, it's not clear to me the distinction among abilities, permissions list and capabilities.

docs/src/src/PERMISSIONS.md
bezerrablockchain: @antomor sorry, I didnt find any SPEC folder, should I create one?

docs/src/src/PERMISSIONS.md
bezerrablockchain: Fixed

docs/src/src/PERMISSIONS.md
bezerrablockchain: Added a new section specific for this contract

docs/src/src/PERMISSIONS.md
bezerrablockchain: Added a new section specific for the `BuilderRegistry`

docs/src/src/PERMISSIONS.md
jmendiola222: Here: https://github.com/RootstockCollective/collective-rewards-sc/tree/main/specs

specs/PERMISSIONS.md
```diff
@@ -0,0 +1,114 @@
+# GovernanceManagerRootstockCollective.sol
+You can find bellow the permissions, roles, and functionalities within the `GovernanceManagerRootstockCollective` contract.
+
+## User Roles and Permissions
+### Governor
+- Who: Defined by the `governor` state variable.
+- Permissions and Capabilities:
+    + Can call `Execute` from contracts implementing the `IChangeContractRootstockCollective` interface through `executeChange`.
+    + View validation of their role using `validateGovernor`.
```
DinizSa, line 9: Being able to call `validateGovernor` is not specific to any role. Should it be here?

specs/PERMISSIONS.md
```diff
@@ -0,0 +1,114 @@
+# GovernanceManagerRootstockCollective.sol
```
DinizSa, line 1: `GovernanceManagerRootstockCollective` is used to hold addresses with permissions, then others contracts use those permissions.
 There are other contracts that use these permissions and are not in this list, like `BackersManagerRootstockCollective`. Should all those contracts also be present in this document?

specs/PERMISSIONS.md
```diff
@@ -0,0 +1,114 @@
...
+    + Triggered when a builder requests a reward receiver replacement.
+- `BuilderRewardReceiverReplacementCancelled(address indexed builder_, address newRewardReceiver_)`
+    + Triggered when a builder cancels a reward receiver replacement request.
+- `BuilderRewardReceiverReplacementApproved(address indexed builder_, address newRewardReceiver_)`
+    + Triggered when a builder's reward receiver replacement is approved.
+- `GaugeCreated(address indexed builder_, address indexed gauge_, address creator_)`
+    + Triggered when a gauge is created for a builder.
+- `BuilderMigrated(address indexed builder_, address indexed migrator_)`
+    + Triggered when a builder is migrated.
+- `NewCycleDurationScheduled(uint256 newCycleDuration_, uint256 cooldownEndTime_)`
```
DinizSa, line 111: is this event relevant here? same applies to other `BuilderRegistryRootstockCollective` events

specs/PERMISSIONS.md
bezerrablockchain: Its just a mention that is possible to verfify if an account is the governor by calling this external function

specs/PERMISSIONS.md
bezerrablockchain: I think is not mandatory, if other people also thinks its not relevant I can remove it.

specs/PERMISSIONS.md
bezerrablockchain: Hummm, maybe a good idea.  My first tough was to cover only the `BuilderRegistryRootstockCollective` as the roles are in fact here, instead of `BackersManagerRootstockCollective`

DinizSa: Sorry, a fix to my comment: `BackersManagerRootstockCollective` does not use permissions.
A better example would be `RewardDistributorRootstockCollective`, that uses the `foundation` address stored in the `GovernanceManagerRootstockCollective` (ie. to set the permissions for the `sendRewards` method).

specs/PERMISSIONS.md
DinizSa: does not seem to fit the section `permissions and capabilities`, that would describe what the role can do. I see methods such as `validateGovernor` as implementation details that do not need to be here. wdyt?

specs/PERMISSIONS.md
DinizSa: I find [velodrome approach](https://github.com/velodrome-finance/contracts/blob/main/PERMISSIONS.md#permissions-list) nice, as they list all the contracts, roles used in that contract, and list of things the role can do in that contract.

bezerrablockchain: Sure, I am adding this.  I am also adding Gauge as well

specs/PERMISSIONS.md
antomor: being the file called permissions, I'd remove the events part since may be out of scope.

specs/PERMISSIONS.md
antomor: if I read it without checking the code, it looks like only the `governor` can call `validateGovernor`.

bezerrablockchain: I rephrased it to express a more clear capability

specs/PERMISSIONS.md
bezerrablockchain: I agree as wekk and as we have a consensus here, I am removing those sections from doc

PR #135, opened 2024-12-20
Code reorg

jmendiola222:
## What

- Makes BackerManager inherit from Interface
- Moves all BackerManager related contract files to `backerManager` folder

## Why

- makes de code more consistence and clear

## Testing

- all test should pass


antomor (approved):
LGTM

PR #137, opened 2024-12-26
chore(docs): remove max foundry version limit

jurajpiar:
## What

- Removes foundry max version requirement from docs (after testing that deployment works with the latest)

## Testing

- deployment to testnet should succeed

## Refs

- https://rsklabs.atlassian.net/browse/TOK-513


jmendiola222 (approved):
Worked for me. :+1:

PR #138, opened 2025-01-03
TOK-553: Split backers manager builder registry

jurajpiar:
## What

- makes builders registry a standalone contract
- Optimize gas costs after contract split

## Why

- (the reason these changes are wanted)

## Testing

- (describe how to test the changes included)

## Refs

https://rsklabs.atlassian.net/browse/TOK-553
https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?assignee=712020%3A42dabdd3-0b32-4060-be21-62cfd46a1751&selectedIssue=TOK-573


src/backersManager/BuilderRegistryRootstockCollective.sol
```diff
@@ -5,15 +5,18 @@ import { CycleTimeKeeperRootstockCollective } from "./CycleTimeKeeperRootstockCo
...
 import { IGovernanceManagerRootstockCollective } from "../interfaces/IGovernanceManagerRootstockCollective.sol";
+import { ICollectiveRewardsCheckRootstockCollective } from
+    "../interfaces/ICollectiveRewardsCheckRootstockCollective.sol"; //def not
 
 /**
  * @title BuilderRegistryRootstockCollective
  * @notice Keeps registers of the builders
  */
-abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstockCollective, ERC165Upgradeable {
+contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstockCollective, ERC165Upgradeable {
```
DinizSa, line 19: Both `BackersManager` and `BuilderRegistry` will need to inherit from UpgradeableRootstockCollective

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -309,18 +296,18 @@ contract BackersManagerRootstockCollective is
...
         uint256 totalPotentialReward_,
         uint256 timeUntilNextCycle_
-    )
-        internal
-        returns (uint256 newbackerTotalAllocation_, uint256 newTotalPotentialReward_)
-    {
+    ) internal returns (uint256 newbackerTotalAllocation_, uint256 newTotalPotentialReward_) {
         // reverts if builder was not activated or approved by the community
-        _validateWhitelisted(gauge_);
+        builderRegistry.validateWhitelisted(gauge_);
```
DinizSa, line 301: a memory variable can be created in the function scope to reduce the one storage read
`address _builderRegistry = builderRegistry`
and replacing with `_builderRegistry` on line 301 and 310

DinizSa: and maybe a ugly but more performant change, would be to pass the builderRegistry address in the function arguments, reducing even more the storage reads when this function is invoked for an array in `allocateBatch`, wdyt?

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -273,7 +256,9 @@ contract BackersManagerRootstockCollective is
     function claimBackerRewards(address rewardToken_, GaugeRootstockCollective[] memory gauges_) external {
         uint256 _length = gauges_.length;
         for (uint256 i = 0; i < _length; i = UtilsLib._uncheckedInc(i)) {
-            if (gaugeToBuilder[gauges_[i]] == address(0)) revert GaugeDoesNotExist();
+            if (builderRegistry.gaugeToBuilder(gauges_[i]) == address(0)) {
```
DinizSa, line 259: a memory variable for builderRegistry can be created outside the loop, reducing the storage reads

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -370,19 +355,19 @@ contract BackersManagerRootstockCollective is
     function _distribute() internal returns (bool) {
         uint256 _newTotalPotentialReward = tempTotalPotentialReward;
         uint256 _gaugeIndex = indexLastGaugeDistributed;
-        uint256 _gaugesLength = getGaugesLength();
+        uint256 _gaugesLength = builderRegistry.getGaugesLength();
```
DinizSa, line 358: a memory variable for builderRegistry can be created at the function level, reducing the storage reads

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -472,22 +461,26 @@ contract BackersManagerRootstockCollective is
...
-    function _resumeGaugeShares(GaugeRootstockCollective gauge_) internal override notInDistributionPeriod {
+    function resumeGaugeShares(GaugeRootstockCollective gauge_) external onlyBuilderRegistry notInDistributionPeriod {
         // gauges cannot be resumed before the distribution,
         // incentives can stay in the gauge because lastUpdateTime > lastTimeRewardApplicable
         if (_periodFinish <= block.timestamp) revert BeforeDistribution();
         // allocations are considered again for the reward's distribution
         // if there was a distribution we need to update the shares with the full cycle duration
-        if (haltedGaugeLastPeriodFinish[gauge_] < _periodFinish) {
-            (uint256 _cycleStart, uint256 _cycleDuration) = getCycleStartAndDuration();
+        if (builderRegistry.haltedGaugeLastPeriodFinish(gauge_) < _periodFinish) {
```
DinizSa, line 470: a memory variable for `builderRegistry` can be created at the function level, reducing the storage reads

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -429,19 +414,23 @@ contract BackersManagerRootstockCollective is
...
+    ) internal returns (uint256) {
         uint256 _rewardShares = gauge_.rewardShares();
         // [N] = [N] * [N] / [N]
         uint256 _amountERC20 = (_rewardShares * rewardsERC20_) / totalPotentialReward_;
         // [N] = [N] * [N] / [N]
         uint256 _amountCoinbase = (_rewardShares * rewardsCoinbase_) / totalPotentialReward_;
-        uint256 _backerRewardPercentage = getRewardPercentageToApply(gaugeToBuilder[gauge_]);
-        return gauge_.notifyRewardAmountAndUpdateShares{ value: _amountCoinbase }(
-            _amountERC20, _backerRewardPercentage, periodFinish_, cycleStart_, cycleDuration_
+        uint256 _backerRewardPercentage = builderRegistry.getRewardPercentageToApply(
```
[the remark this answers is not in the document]
jurajpiar: sounds good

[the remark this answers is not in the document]
DinizSa: just to document, I decided to NOT apply this since `getRewardPercentageToApplyByGauge` would need to to do an extra storage read, not making the small gas savings worth the extra code imo.

gas_comparison.csv
antomor, line 1: @DinizSa , I guess we could delete this before merging.

.gas-snapshot
antomor, line 1: @DinizSa I guess we can delete this once the PR is merged

script/Deploy.s.sol
```diff
@@ -70,19 +72,31 @@ contract Deploy is Broadcaster, OutputWriter {
...
-            address(_gaugeFactory),
-            address(_rewardDistributorProxy),
             _cycleDuration,
             _cycleStartOffset,
-            _distributionDuration,
-            _rewardPercentageCooldown
+            _distributionDuration
         );
         saveWithProxy("BackersManagerRootstockCollective", address(_backersManagerImpl), address(_backersManagerProxy));
+        _builderRegistryProxy.setBackersManager(_backersManagerProxy);
```
antomor, line 99: I think that this needs to be called after `vm.broadcast()` if we want it to take effect

src/backersManager/BuilderRegistryRootstockCollective.sol
```diff
@@ -116,46 +124,59 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
...
-            governanceManager_, cycleDuration_, cycleStartOffset_, distributionDuration_
-        );
-        __ERC165_init();
+        __Upgradeable_init(governanceManager_);
         gaugeFactory = GaugeFactoryRootstockCollective(gaugeFactory_);
         rewardDistributor = rewardDistributor_;
         rewardPercentageCooldown = rewardPercentageCooldown_;
     }
 
+    function setBackersManager(BackersManagerRootstockCollective backersManager_) external {
```
antomor, line 160: Does it make sense to have a function to initialise it (as it is now) and another to set it (in case the backers manager address change it)? Or will we just upgrade it and we keep it for initialisation only? If so, I'd rename it to `initialiseBackersManagerAddress` to make it clearer.

src/backersManager/BuilderRegistryRootstockCollective.sol
```diff
@@ -302,7 +323,7 @@ abstract contract BuilderRegistryRootstockCollective is CycleTimeKeeperRootstock
         builderState[builder_].communityApproved = false;
 
         _haltGauge(_gauge);
-        _rewardTokenApprove(address(_gauge), 0);
+        backersManager.rewardTokenApprove(address(_gauge), 0);
```
antomor, line 326: Does it make sense to have a utility for this?

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -56,6 +62,13 @@ contract BackersManagerRootstockCollective is
         _;
     }
 
+    modifier onlyBuilderRegistry() {
+        if (msg.sender != address(builderRegistry)) {
+            revert NotAuthorised();
```
antomor, line 67: I've just noticed that in some parts we used `Authorized` while in others we use `Authorised`. I think we should be consistent.

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -165,15 +173,16 @@ contract BackersManagerRootstockCollective is
...
     function allocate(GaugeRootstockCollective gauge_, uint256 allocation_) external notInDistributionPeriod {
-        (uint256 _newbackerTotalAllocation, uint256 _newTotalPotentialReward) = _allocate(
+        (uint256 _newBackerTotalAllocation, uint256 _newTotalPotentialReward) = _allocate(
             gauge_,
             allocation_,
             backerTotalAllocation[msg.sender],
             totalPotentialReward,
-            timeUntilNextCycle(block.timestamp)
+            timeUntilNextCycle(block.timestamp),
+            builderRegistry
```
antomor, line 182: Why we pass it here rather than using the field where needed?

src/backersManager/BackersManagerRootstockCollective.sol
antomor, line 1: @DinizSa just a question: have you also considered or tried to move the gauges in the Builder registry?

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -77,7 +77,9 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
     /// @notice address of the token rewarded to builder and voters
     address public rewardToken;
     /// @notice BackersManagerRootstockCollective contract address
-    IBackersManagerRootstockCollective public backersManager;
+    BackersManagerRootstockCollective public backersManager;
+    /// @notice BuilderRegistryRootstockCollective contract address
+    BuilderRegistryRootstockCollective public builderRegistry;
```
antomor, line 82: Any specific reason for not using the interfaces?

src/backersManager/BuilderRegistryRootstockCollective.sol
DinizSa: is a good point, I don't think it needs to be changed after, as it is already known at deployment time and will not change. I will remove this function and add it in the initialize function

src/backersManager/BuilderRegistryRootstockCollective.sol
DinizSa: wdym? to add that code to an internal function?

src/backersManager/BackersManagerRootstockCollective.sol
DinizSa: agree, I went with the US version `Authorized`

src/gauge/GaugeRootstockCollective.sol
DinizSa: Using the smart contract directly has the same deployment cost and gas costs. 
I could create an interface `IBuilderRegistryRootstockCollective`, then I would need to create interfaces for all our contracts, which I'm not sure is the direction we want to go.

src/backersManager/BuilderRegistryRootstockCollective.sol
antomor: either an internal function or a library. What do you think it's better?

src/gauge/GaugeRootstockCollective.sol
antomor: ok, got it, thank you for the explanation

src/backersManager/BuilderRegistryRootstockCollective.sol
antomor: I made a bit of confusion here 😅

src/backersManager/BuilderRegistryRootstockCollective.sol
DinizSa: @antomor I agree that initialize makes more sense for this situation, but initialize is a "reserved" word in the upgradeables contracts patterns that gives the impression it will be called after the contract deploy/upgrade. 
For this reason I prefer setup, is that ok?

script/Deploy.s.sol
antomor: I think we can execute both `setBackersManager` after one single vm.broadcast. Could we try it, please?

src/backersManager/BuilderRegistryRootstockCollective.sol
antomor: In the `BuilderRegistry` we used `initializeBackersManager`. I'd be consistent and use the same unless you've a reason for changing it.

script/Deploy.s.sol
DinizSa: not sure I get your point. Each `vm.broadcast` only applies to the next transaction.

src/backersManager/BuilderRegistryRootstockCollective.sol
DinizSa: you mean `_rewardDistributorProxy.initializeCollectiveRewardsAddresses` instead? since we already have it I'm ok to keep that consistency.

antomor (approved):

PR #139, opened 2025-01-06
TOK-553: [Analysis] Make gauge track builder

jurajpiar:
## What

- removes gaugeToBuilder from BuilderRegistry
- adds builder to Gauge

## Why

- to reduce deloyment bytesize of the BackersManager contract

## Testing

- `forge snapshot` for an idea of tx execution cost 
- `forge build --sizes` for bytecode size

## Refs

https://rsklabs.atlassian.net/browse/TOK-553


antomor:
we decided not to proceed with this strategy

PR #141, opened 2025-01-09
TOK-342: gas optimization

antomor:
## What

- change visibility of multiple functions from public to external and added logic to reduce gas

## Why

- to reduce gas cost


src/gauge/GaugeRootstockCollective.sol
```diff
@@ -329,32 +329,36 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
         onlyBackersManager
         returns (uint256 allocationDeviation_, uint256 rewardSharesDeviation_, bool isNegative_)
     {
+        address _rewardToken = rewardToken;
```
DinizSa, line 332: this change increases slightly the gas (1 more memory variable declaration - 3 gas) cost if totalAllocation is not 0, and decreases it more significantly (1 less storage read - 2,100 gas) if totalAllocation is 0.
Given the gauge total allocation is rarely 0, this change might not be bring benefit. wdyt?

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -329,32 +329,36 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
...
-            rewardShares += rewardSharesDeviation_;
-        } else {
-            allocationDeviation_ = _previousAllocation - allocation_;
-            // avoid underflow because rewardShares may not be correctly updated if the distribution was skipped
-            rewardSharesDeviation_ = Math.min(rewardShares, allocationDeviation_ * timeUntilNextCycle_);
-            totalAllocation -= allocationDeviation_;
-            rewardShares -= rewardSharesDeviation_;
-            isNegative_ = true;
+        
+        unchecked {
```
DinizSa, line 344: 1) Just curiosity, is the gas savings significant?
2) Can you please add a comment on why is it safe to perform these opertions unchecked?

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -430,24 +434,27 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
...
-            _backerAmountCoinbase,
-            periodFinish_,
-            _timeUntilNextCycle
-        );
+        address rewardToken_ = rewardToken;
+        unchecked {
+            uint256 _backerAmountERC20 = UtilsLib._mulPrec(backerRewardPercentage_, amountERC20_);
+            uint256 _backerAmountCoinbase = UtilsLib._mulPrec(backerRewardPercentage_, msg.value);
+            uint256 _timeUntilNextCycle = UtilsLib._calcTimeUntilNextCycle(cycleStart_, cycleDuration_, block.timestamp);
+            _notifyRewardAmount(
```
DinizSa, line 442: unchecked has no effect on wrapped functions, can be excluded

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -512,9 +519,9 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
      *  address(uint160(uint256(keccak256("COINBASE_ADDRESS")))) is used for coinbase address
      */
     function _left(address rewardToken_) internal view returns (uint256) {
-        RewardData storage _rewardData = rewardData[rewardToken_];
+        uint256 _rewardRate = rewardData[rewardToken_].rewardRate;
         // [N] = ([N] - [N]) * [PREC] / [PREC]
-        return UtilsLib._mulPrec(backersManager.periodFinish() - block.timestamp, _rewardData.rewardRate);
+        return UtilsLib._mulPrec(backersManager.periodFinish() - block.timestamp, _rewardRate);
```
DinizSa, line 524: 👍  we can even avoid the memory variable declaration and do it as:
return UtilsLib._mulPrec(backersManager.periodFinish() - block.timestamp, rewardData[rewardToken_].rewardRate);

src/gauge/GaugeRootstockCollective.sol
bezerrablockchain: hi @DinizSa
the changes I proposed here is based only on the before and after test-profile comparison.  In this case, wo the variable creation the gas-reporter show as bellow:
before:
| Function Name                                                 | min             | avg     | median | max        | # calls |
| notifyRewardAmountAndUpdateShares          | 2547            | 68805 | 86550  | 125316   | 57      |
after:
| Function Name                                                 | min             | avg     | median | max       | # calls |
| notifyRewardAmountAndUpdateShares          | 2547            | 68686 | 86428   | 125194  | 57      |

maybe coz the operation of create a new memory variable + read this variable twice from memory is requiring less gas than reading the state variable twice.  wdyt ?

src/gauge/GaugeRootstockCollective.sol
bezerrablockchain: yes, I kept those in the same unchecked block coz they are a variable declaration and it would not break the code readability if they are all in the same block, as you said, theres no gain for the wrapped function, the unchecked is acting on top of the minus operation inside the block

src/gauge/GaugeRootstockCollective.sol
bezerrablockchain: imo many of this proposed changes does not have the significant gas saving.  But we have a document [here](https://www.notion.so/iovlabs/TOK-342-Gas-optimization-157c132873f980b2a97accc370f39b3c) explaining about the changes.

src/gauge/GaugeRootstockCollective.sol
bezerrablockchain: tks.  I'll change it

src/gauge/GaugeRootstockCollective.sol
DinizSa: yeah understood, they need to be inside so the variables exist in the function scope. 👍

src/gauge/GaugeRootstockCollective.sol
DinizSa: 1. I did a quick search and seems that unchecked saves around 19 gas per math operation

if an overflow happens in here there will not be a bug instead of an error thrown. Are we sure this is completely safe to not check overflows in all these operations?

DinizSa: if yes, a comment in the code explaining the reasoning would be useful for future developers that might not know about the documentation.

bezerrablockchain: I cant see a real scenario where a over/under flow may happens here, ofc one uint256 cant accommodate a sum of two maximum uint256, but are those scenario real here?  Thats was my reasoning.

But I totally agree that this trade-off does not worth... I am removing the unchecked.

src/gauge/GaugeRootstockCollective.sol
DinizSa: > maybe coz the operation of create a new memory variable + read this variable twice from memory is requiring less gas than reading the state variable twice. wdyt ?
yes, but when `if (totalAllocation == 0) {` is true (line 336) the read from the storage only happens 1 time, making this change potentially inefficient, since gauge.totalAllocation will 0 only until the first allocation for that gauge happens.

bezerrablockchain: ok I got it!  So the tests-profile results may show some gas gain coz the most of test scenarios start with `totalAllocation == 0` but in the real use cases the most of operations will be done with it <> 0.

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -430,24 +431,27 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
...
-        );
-        _notifyRewardAmount(
-            UtilsLib._COINBASE_ADDRESS,
-            msg.value - _backerAmountCoinbase,
-            _backerAmountCoinbase,
-            periodFinish_,
-            _timeUntilNextCycle
-        );
+        address _rewardToken = rewardToken;
+        unchecked {
```
DinizSa, line 435: I feel hesitant on adding unchecked to such operations, for the reasons stated in the other thread. Is a trade off between security and gas, and not that much gas is saved.

bezerrablockchain: removed in favor of security

PR #144, opened 2025-01-14
TOK-490/deadlock

franciscotobar:
## What

- Allow distributing without allocations

## Why

- Division error if there are builders and the total allocation is 0

## Refs

- [TOK-490](https://rsklabs.atlassian.net/browse/TOK-490)

### DIFF
--- a/src/backersManager/BackersManagerRootstockCollective.sol
+++ b/src/backersManager/BackersManagerRootstockCollective.sol
@@ -379,6 +379,13 @@ contract BackersManagerRootstockCollective is
         uint256 _totalPotentialReward = totalPotentialReward;
         uint256 __periodFinish = _periodFinish;
         (uint256 _cycleStart, uint256 _cycleDuration) = getCycleStartAndDuration();
+
+        // no rewards to distribute since there are no allocations
+        if (_totalPotentialReward == 0) {
+            _finishDistribution();
+            return true;
+        }
+
         // loop through all pending distributions
         while (_gaugeIndex < _lastDistribution) {
             _newTotalPotentialReward += _gaugeDistribute(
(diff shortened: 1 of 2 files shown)


bezerrablockchain (approved):
LGTM

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -393,23 +400,28 @@ contract BackersManagerRootstockCollective is
...
         if (_lastDistribution == _gaugesLength) {
-            emit RewardDistributionFinished(msg.sender);
-            indexLastGaugeDistributed = 0;
-            rewardsERC20 = rewardsCoinbase = 0;
-            onDistributionPeriod = false;
-            tempTotalPotentialReward = 0;
+            _finishDistribution();
             totalPotentialReward = _newTotalPotentialReward;
-            _periodFinish = cycleNext(block.timestamp);
+            rewardsERC20 = rewardsCoinbase = 0;
```
antomor, line 408: NIT: Why don't we reset `rewardsERC20` and `rewardsCoinbase` in `_finishDistribution` also?

test/BackersManagerRootstockCollective.t.sol
antomor, line 1: @franciscotobar can we add a test here where the distribution was successful first with some allocations and then it's also successful when the allocations are removed?

antomor (changes requested):
@franciscotobar The changes look great, thank you. I'd just add an additional test.

src/backersManager/BackersManagerRootstockCollective.sol
antomor: Got it, thank you for the clarification.

PR #146, opened 2025-01-20
TOK-338: nft boost poc

DinizSa:
## What
This is a POC of a possible approach to implement the NFT boosting.
An off-chain component is needed to call `updateBackerNftBoost` for each backer that has allocations and a NFT.

`updateBackerNftBoost` function changes the backer rewards by applying a boost
   - On NFT campaign start: call this function for each backer that has allocations AND a NFT, with the NFT boost percentage
   - On NFT campaign end: call this function for each backer previously boosted, passing NFT boost percentage as 0
   - It does not increase the backer stRif allocated to the system, allowing the backer to withdraw the expected stRif
  
Note POC: for simplicity no safety checks are performed (ie. backer owns the NFT, does not apply, etc)

## Why

- See if the core of this approach works

## Testing

- Unit test

## Refs

- [Jira 338 ticket](https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?selectedIssue=TOK-338&text=nft)

### DIFF
--- a/src/backersManager/BackersManagerRootstockCollective.sol
+++ b/src/backersManager/BackersManagerRootstockCollective.sol
@@ -97,49 +97,43 @@ contract BackersManagerRootstockCollective is
     /// @notice address of the builder registry contract
     BuilderRegistryRootstockCollective public builderRegistry;
 
-    address public nftContractBoost;
-    uint256 public nftContractBoostPct; // 1 = 1%, 100 = 100%
-    mapping(address nftContract => mapping(address backer => bool hasNftBoost)) public hasNftBoost;
+    mapping(address backer => uint256 boostPct) public backerBoostPct; // 1 = 1%, 100 = 100%
 
     /**
-     * @dev This function will set the NFT contract address and the boost percentage
-     * @dev setNftContract should be called before calling updateBackerNftBoost, both when the NFT campaign starts or
-     * finishes
-     * @dev POC: for simplicity, only one NFT contract can be set at a time, and no safety checks are performed
-     */
-    function setNftContract(address nftContractBoost_, uint256 boostPct_) external onlyValidChangerOrFoundation {
-        governanceManager.validateAuthorizedChanger(msg.sender);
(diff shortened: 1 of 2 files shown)


src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -97,6 +97,45 @@ contract BackersManagerRootstockCollective is
...
+    {
+        governanceManager.validateAuthorizedChanger(msg.sender);
+        require(backerBoostPct_ <= 100, "invalid boost percentage");
+
+        uint256 _currentBackerBoostPct = backerBoostPct[backer_];
+        for (uint256 i = 0; i < allocatedGauges_.length; i++) {
+            uint256 _allocationOfBacker = allocatedGauges_[i].allocationOf(backer_);
+            if (_allocationOfBacker == 0) continue;
+            // uint256 _gaugeBoostedAllocation = _allocationOfBacker + _allocationOfBacker * nftContractBoostPct / 100;
+            uint256 _gaugeNewBoostedAllocation =
```
antomor, line 129: Why are we dividing by `(100 + _currentBackerBoostPct)` instead of just `100`

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -97,6 +97,45 @@ contract BackersManagerRootstockCollective is
...
+                (100 + backerBoostPct_) * _allocationOfBacker / (100 + _currentBackerBoostPct);
+
+            (, uint256 _rewardSharesDeviation,) =
+                allocatedGauges_[i].allocate(backer_, _gaugeNewBoostedAllocation, timeUntilNextCycle(block.timestamp));
+            totalPotentialReward += _rewardSharesDeviation;
+        }
+        backerBoostPct[backer_] = backerBoostPct_;
```
antomor, line 136: The current solution supports one NFT campaign at a time.

DinizSa:
following an offchain approach for now, we might get back to this approach later.

PR #147, opened 2025-01-28
QA: update sc addresses

antomor:
## What

- update the contract addresses for QA

## Why

- to allow them to test the new version of the contract


antomor:
We don't need to merge this since we migrated the contracts.

PR #148, opened 2025-01-29
TOK-602: review why it is failing

antomor:
## What

- Fix the hardhat artifact creation

## Why

- Right now the hardhat artifact aren't created

## How to test it

- `bun run deploy` (after configuring the environment with `.chain_id`)
  - the expected output is to have in the folder configured with `.chain_id`
    - `contract_addresses.json`
    - one file for each contract deployed and the file should be named with the following format: `<contract_name>.abi.json`


.env.example
```diff
@@ -15,7 +15,8 @@ export CYCLE_START_OFFSET=0
 # Optional
 export FOUNDRY_PROFILE="default"
 export ACCOUNT="0xYOUR_ADDRESS"
-export OMIT_HARDHAT_ARTIFACTS=false
+export OMIT_ABI_COPY=false
+export NO_BROADCAST=false
```
DinizSa, line 19: just to understand, is the intent of this variable to be used to run and validate the scripts, without doing changes in the network?

franciscotobar (approved):
LGTM

PR #151, opened 2025-01-31
[POC] refactor(builder_state): as state machine

jurajpiar:
## What

- POC State machine approach to storing builder state

## Why

- gas cost + contract size + complexity

- [Analysis task](https://rsklabs.atlassian.net/browse/TOK-576)
- [Diagrams](https://miro.com/app/board/uXjVKa78bEM=/?moveToWidget=3458764620862033337&cot=14)

### DIFF
--- a/src/builderRegistry/BuilderRegistryRootstockCollective.sol
+++ b/src/builderRegistry/BuilderRegistryRootstockCollective.sol
@@ -23,13 +23,13 @@ interface StateMachineErrors {
 }
 
 interface BuilderEvents {
-    event BuilderStateChange(address indexed builder_, BuilderState prevState_, BuilderState newState_);
-    event BuilderActivated(address indexed builder_, address rewardReceiver_, uint64 rewardPercentage_);
-    event Dewhitelisted(address indexed builder_);
+    event BuilderStateChanged(address indexed builder_, BuilderState prevState_, BuilderState newState_);
+    event RewardsInitialized(address indexed builder_, address rewardReceiver_, uint64 rewardPercentage_);
+    event Blacklisted(address indexed builder_);
     event Paused(address indexed builder_, bytes20 reason_);
     event Unpaused(address indexed builder_);
-    event Revoked(address indexed builder_);
-    event Permitted(address indexed builder_, uint256 rewardPercentage_, uint256 cooldown_);
+    event ReceivingRewardsPaused(address indexed builder_);
+    event ReceivingRewardsResumed(address indexed builder_, uint256 rewardPercentage_, uint256 cooldown_);
     event BackerRewardPercentageUpdateScheduled(address indexed builder_, uint256 rewardPercentage_, uint256 cooldown_);
     event BuilderRewardReceiverReplacementRequested(address indexed builder_, address newRewardReceiver_);
     event BuilderRewardReceiverReplacementCancelled(address indexed builder_, address newRewardReceiver_);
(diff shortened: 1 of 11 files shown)


src/builderRegistry/BuilderRegistryRootstockCollective.sol
```diff
@@ -243,32 +244,49 @@ contract BuilderRegistryRootstockCollective is UpgradeableRootstockCollective {
...
+         *             DEAD = 5
+         *         }
+         *
+         *         0 -> 2 Inactive -> KYCApproved
+         *         1 -> 3 CommunityApproved -> Active
+         */
+        BuilderState _state = builderState[builder_];
+        BuilderState _targetState = BuilderState(uint8(_state) + 2);
+
+        if (backerRewardPercentage[builder_].cooldownEndTime == 0) {
```
DinizSa, line 274: We are using `cooldownEndTime` to check if a builder ever exists.
This seems a hidden coupling that might be error prone. for example if `cooldownEndTime` is no longer used, or some logic is added that sets it to 0, this flow would break. 
The issue seems to arise because activating a builder does not change the builder state.

src/builderRegistry/BuilderRegistryRootstockCollective.sol
```diff
@@ -9,18 +9,36 @@ import { GaugeFactoryRootstockCollective } from "../gauge/GaugeFactoryRootstockC
...
- */
-contract BuilderRegistryRootstockCollective is UpgradeableRootstockCollective {
-    using EnumerableSet for EnumerableSet.AddressSet;
+enum BuilderState {
+    Inactive,
+    CommunityApproved,
+    KYCApproved,
+    Active,
+    Paused,
+    DEAD // Marks the end for iterations over State
```
DinizSa, line 18: Couple of observations:
1. DEAD state is being used as a builder state when he is dewhitelisted. but by the comment and capital case, it is a special state that is never a valid builder state.
2. If we need in the future to add more states, by having DEAD as the last state, is currently limiting how many states we can have.

An alternative would be to move this DEAD state to be the first state and rename it to inexistent, or similar. since by default all addresses will have the enum default value (0), this would also make sense for all non builders addresses. It would also not limit new states additions in the future.

jurajpiar: that was my first approach, but the thing is that when an address gets community revoked, it can never return, which is why it cannot be 0th state.
the DEAD fits this well, as it is never used as a state as such, so it's always invalid (DEAD), as is any address reaching this state (in the context ofc).

as for the future upgrades, I think that any such updates should be breaking anyway as it would imply the change in the state machine itself. Any non-state machine flags we might want to add should not be a part of the state itself.

src/builderRegistry/BuilderRegistryRootstockCollective.sol
jurajpiar: I get your point that this mechanism should be more stable in that regard, but, surely, if some logic is considered to be added that touches any existing data at all, it should consider the consequences it has on the given data.
Cooldown on the reward percentage also seems like such an important feature that I'm not sure it would be easily removed.

But more to the point, would you offer an alternative way to detect activation?

DinizSa: my suggestion would be for the builder state to be updated once `_activateBuilder` is called, by adding a new `activated` or `initialised` state. I think this makes sense as that address permissions in the system are different after that method is executed. it feels strange that a random address has the same state as an activated address.

src/builderRegistry/BuilderRegistryRootstockCollective.sol
DinizSa: renaming DEAD state to communityRevoked (or simillar), and not impose this state to be the last state of all the states, might solve this, wdyt?

antomor:
We decided not to implement it now due to the impact it will have, but we'll reconsider it when new business needs will arise.

PR #155, opened 2025-02-12
Deployment qa v1

DinizSa:
## What

- fresh v1 deploy in testnet for qa environment
- update env to match notion env config page

## Why

- previous SC got in a situation where they could not get updated, due to a buggy migration contract


antomor (approved):

antomor:
@DinizSa I don't think we need to merge this PR, since we could directly merge the updated ones.

PR #190, opened 2025-04-10
Tok 679/erc20 approve not validated

Hakob23:
## What

Handling the scenarios of erc20 token returning false.

## Why

In most cases, this will not happen, as almost all erc20 standard tokens have their built-in reversion logic, but still, if the given token returns false on the approval, the contract will revert with the corresponding message.

## Testing

- There are not any separate unit tests written for this function as it is used by other contracts' functions that are being tested.

### DIFF
--- a/src/backersManager/BackersManagerRootstockCollective.sol
+++ b/src/backersManager/BackersManagerRootstockCollective.sol
@@ -41,6 +41,7 @@ contract BackersManagerRootstockCollective is
     error AlreadyOptedInRewards();
     error BackerHasAllocations();
     error InvalidAddress();
+    error RewardTokenNotApproved();
 
     // -----------------------------
     // ----------- Events ----------
@@ -565,11 +566,15 @@ contract BackersManagerRootstockCollective is
     /**
      * @notice approves rewardTokens to a given gauge
      * @dev give full allowance when it is community approved and remove it when it is dewhitelisted
+     * reverts if the reward token contract returns false on the approval
      * @param gauge_ gauge contract to approve rewardTokens
      * @param value_ amount of rewardTokens to approve
      */
     function rewardTokenApprove(address gauge_, uint256 value_) external onlyBuilderRegistry {
-        IERC20(rewardToken).approve(gauge_, value_);
+        if(IERC20(rewardToken).approve(gauge_, value_)) {
+            revert RewardTokenNotApproved();
+        }
+        
     }
 
     /**
--- a/src/backersManager/BackersManagerRootstockCollective.sol
+++ b/src/backersManager/BackersManagerRootstockCollective.sol
@@ -571,7 +571,7 @@ contract BackersManagerRootstockCollective is
      * @param value_ amount of rewardTokens to approve
      */
     function rewardTokenApprove(address gauge_, uint256 value_) external onlyBuilderRegistry {
-        if(IERC20(rewardToken).approve(gauge_, value_)) {
+        if(!IERC20(rewardToken).approve(gauge_, value_)) {
             revert RewardTokenNotApproved();
         }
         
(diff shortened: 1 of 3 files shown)


src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -565,11 +566,15 @@ contract BackersManagerRootstockCollective is
...
      * @notice approves rewardTokens to a given gauge
      * @dev give full allowance when it is community approved and remove it when it is dewhitelisted
+     * reverts if the reward token contract returns false on the approval
      * @param gauge_ gauge contract to approve rewardTokens
      * @param value_ amount of rewardTokens to approve
      */
     function rewardTokenApprove(address gauge_, uint256 value_) external onlyBuilderRegistry {
-        IERC20(rewardToken).approve(gauge_, value_);
+        if(!IERC20(rewardToken).approve(gauge_, value_)) {
+            revert RewardTokenNotApproved();
```
DinizSa, line 575: we could add a test with a mock IERC20 contract to test this condition, do you think is worth it?

DinizSa (approved):
LGTM, just left a comment for a possible test case

src/backersManager/BackersManagerRootstockCollective.sol
Hakob23: Yes, I have already added it. However, introducing an additional mockErc20 would definitely be an overkill, as the deployment mechanism is pretty complex, and we had to deploy every contract one more time with that new token in order not to disturb the overall deployment architecture, which, for just one test was not worth it, so I've overrode the approve function for the existing MockErc20 contract.

test/BackersManagerRootstockCollective.t.sol
```diff
@@ -100,6 +100,23 @@ contract BackersManagerRootstockCollectiveTest is BaseTest {
...
+    /**
+     * SCENARIO: should revert if the reward token approval returns false
+     * @dev during the initial setup, the contract was already approved for a spending allowance greater than zero,
+     * and the mock reward token implements approve() logic which forces users to negate the allowance before approving a new quantity of allowance
+     * The function will revert if called before negating the previous allowance
+     */
+    function test_RevertRewardTokenApprove() public {
+        // Should not revert, as the allowance is negated before approving a new amount
+        vm.startPrank(address(builderRegistry));
+        backersManager.rewardTokenApprove(address(gauge), 0);
```
antomor, line 112: Do we need this if we approve the max in the next line?

test/BackersManagerRootstockCollective.t.sol
```diff
@@ -100,6 +100,23 @@ contract BackersManagerRootstockCollectiveTest is BaseTest {
...
+        backersManager.rewardTokenApprove(address(gauge), type(uint256).max);
+
+        // Should revert, as the allowance is not negated before approving a new amount
+        vm.expectRevert(BackersManagerRootstockCollective.RewardTokenNotApproved.selector);
```
antomor, line 116:
```suggestion
vm.expectRevert(BackersManagerRootstockCollective.RewardTokenNotApproved.selector);
```

test/BackersManagerRootstockCollective.t.sol
Hakob23: Yes, according to the new approve logic, we must make the allowance 0, then give a new approval. This is a legacy defense method against racimg conditions, when the user might give more allowance without knowing about it.
I have covered both successfull and revertion cases in this test. 
Also checked other tests that we have now. They are all compatible with the new approve function.

test/BackersManagerRootstockCollective.t.sol
antomor: I've just removed the mpty line here

antomor (approved):
@Hakob23 Thanks for the changes, LGTM

raullaprida (commented):
lgtm

PR #193, opened 2025-04-10
TOK-680/outdated-dependencies

franciscotobar:
## What

- Update dependencies
- Update `eslint` configuration files

## Why

- Outdated dependencies
- After updating dependencies, the `eslint` configuration files were deprecated
## Testing

- bun run lint

## Refs

- [TOK-680](https://rsklabs.atlassian.net/browse/TOK-680)


antomor (changes requested):
@franciscotobar thanks for the changes. I see that we're changing some eslint configuration, but we still have some outdated dependencies that you can check by running `npm audit`.

script/compare_gas.js
```diff
@@ -18,7 +18,7 @@ function readGasValues(content) {
 }
 
 // Get the current commit hash
-const currentCommit = execSync('git rev-parse HEAD').toString().trim();
+//const currentCommit = execSync('git rev-parse HEAD').toString().trim();
```
antomor, line 21: Should we uncomment this?

script/compare_gas.js
```diff
@@ -1,13 +1,13 @@
 import { execSync } from 'child_process';
 import { readFileSync, writeFileSync } from 'fs';
-import { parse } from 'path';
+//import { parse } from 'path';
```
antomor, line 3: if we don't need this, I think we could remove it completely

antomor (approved):
Apart from some minor suggestions, LGTM

script/compare_gas.js
franciscotobar: I removed it, since it's not necessary

script/compare_gas.js
franciscotobar: I removed it, since it's not necessary

raullaprida (commented):
LGTM

PR #211, opened 2025-04-21
TOK-686(KS-RCF-O-01): best secure code practice

Hakob23:
## What

- Several comments improvements as recommended per Kudelsky review, making `maxDistributionsPerBatch` a storage variable

## Why

- Depending on what max batch size will work out the best, it's better to have the ability to change it after the deployment.

## Testing

- Added `test_RevertUpdateMaxDistributionsPerBatchNotUpgrader()` and `test_UpdateMaxDistributionsPerBatch()` unit tests


src/RewardDistributorRootstockCollective.sol
```diff
@@ -75,7 +75,6 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective
     /**
      * @notice sends rewards to backersManager contract to be distributed to the gauges
      * @dev reverts if is not called by foundation treasury address
-     *  reverts if rewards balance is insufficient
```
DinizSa, line 78: if the funds in the reward distributor are insufficient this function will revert. Should we leave the error message and be more specific? for example
`@dev Reverts if the RewardDistributor's reward's tokens balance are insufficient.`

src/RewardDistributorRootstockCollective.sol
```diff
@@ -86,7 +85,6 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective
     /**
      * @notice sends rewards to backersManager contract and starts the distribution to the gauges
      * @dev reverts if is not called by foundation treasury address
-     *  reverts if rewards balance is insufficient
```
DinizSa, line 89: same as [here](https://github.com/RootstockCollective/collective-rewards-sc/pull/211/files#r2053016887)

test/BackersManagerRootstockCollective.t.sol
```diff
@@ -1815,4 +1815,32 @@ contract BackersManagerRootstockCollectiveTest is BaseTest {
...
+     */
+    function test_RevertUpdateMaxDistributionsPerBatchNotUpgrader() public {
+        // GIVEN a new maxDistributionsPerBatch value
+        uint256 newMaxDistributionsPerBatch = 30;
+
+        // WHEN a non-upgrader tries to update maxDistributionsPerBatch
+        vm.prank(alice);
+        vm.expectRevert(BackersManagerRootstockCollective.NotAuthorized.selector);
+        backersManager.updateMaxDistributionsPerBatch(newMaxDistributionsPerBatch);
+    }
```
DinizSa, line 1845: can we also cover the test case where there are more gauges than `maxDistributionsPerBatch`, to ensure that this parameter works as expected?

src/RewardDistributorRootstockCollective.sol
Hakob23: In the report it was mentioned that in the contract itself it was not reverting anything
`According to the comment, the insufficient reward balance should be reverted, but not
checked in the code. `
So they recommended to remove it. I also think that it should not be there as it is not directly part of the contract's functionality, and the description of the reverting in case of insufficient balance can be found in ERC20 contract when looking inside the function call stack, so it will be a duplicate information. Imo, I don't think we should keep it.

test/BackersManagerRootstockCollective.t.sol
Hakob23: Yes, the test case is added as suggested

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -176,6 +180,14 @@ contract BackersManagerRootstockCollective is
     // used to initialize `builderRegistry` during an upgrade to version 2.
     // The function has been removed since the upgrade was already executed and it's no longer necessary.
 
+    /**
+     * @notice contract version 3 initializer
+     * @param maxDistributionsPerBatch_ maximum number of distributions allowed per batch
+     */
+    function initializeV3(uint256 maxDistributionsPerBatch_) external reinitializer(3) {
```
DinizSa, line 187: This method execution is needed for a correct contract setup. We need to make sure the deployment scripts are calling this method, setting a proper state for the contracts after deployment. maybe reading from an environment variable, and passing it to this function.

antomor: We realised it's not possible to do it during an upgrade, so we'll move forward with the `initializeV3` function

antomor (approved):
LGTM, thanks @Hakob23

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -395,6 +405,21 @@ contract BackersManagerRootstockCollective is
...
     }
 
+    /**
+     * @notice Updates the maximum number of distributions allowed per batch.
+     * @dev reverts if not called by the upgrader
+     * @dev permission will be delegated from upgrader to different role once the GovernanceManagerRootStockCollective
+     * contract is upgraded
+     */
+    function updateMaxDistributionsPerBatch(uint256 maxDistributionsPerBatch_) external {
+        if (msg.sender != governanceManager.upgrader()) {
```
raullaprida, line 415: Not a really important comment, but, you chose to not use a modifier because the restriction won't be used anywhere else right?

raullaprida (commented):
Max distribution batch change LGTM

PR #212, opened 2025-04-22
TOK-700/request-multiple-gauges

franciscotobar:
## What

- Reduce amount of gauge external calls

## Why

- Reduce gas cost

## Gas comparison
- Old
Screenshot 2025-04-28 at 12 05 08 PM

- New
Screenshot 2025-04-28 at 12 01 29 PM

## Refs

- [TOK-700](https://rsklabs.atlassian.net/browse/TOK-700)

### DIFF
--- a/src/builderRegistry/BuilderRegistryRootstockCollective.sol
+++ b/src/builderRegistry/BuilderRegistryRootstockCollective.sol
@@ -517,6 +517,15 @@ contract BuilderRegistryRootstockCollective is UpgradeableRootstockCollective {
         return _gauges.at(index_);
     }
 
+    function getGaugesInRange(uint256 start_, uint256 end_) public view returns (address[] memory gauges_) {
+        uint256 _gaugesLength = _gauges.length();
+        if (start_ >= _gaugesLength || end_ > _gaugesLength || start_ >= end_) revert InvalidAddress();
+        gauges_ = new address[](end_ - start_);
+        for (uint256 i = start_; i < end_; i++) {
+            gauges_[i - start_] = _gauges.at(i);
+        }
+    }
+
     /**
      * @notice return true is gauge is rewarded
      */
--- a/src/builderRegistry/BuilderRegistryRootstockCollective.sol
+++ b/src/builderRegistry/BuilderRegistryRootstockCollective.sol
@@ -39,6 +39,7 @@ contract BuilderRegistryRootstockCollective is UpgradeableRootstockCollective {
     error NotAuthorized();
     error ZeroAddressNotAllowed();
     error BuilderRewardsLocked();
+    error InvalidIndex();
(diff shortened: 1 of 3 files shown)


src/builderRegistry/BuilderRegistryRootstockCollective.sol
```diff
@@ -529,6 +529,15 @@ contract BuilderRegistryRootstockCollective is UpgradeableRootstockCollective {
         return _gauges.at(index_);
     }
 
+    function getGaugesInRange(uint256 start_, uint256 end_) public view returns (address[] memory gauges_) {
+        uint256 _gaugesLength = _gauges.length();
+        if (start_ >= _gaugesLength || end_ > _gaugesLength || start_ >= end_) revert InvalidAddress();
```
DinizSa, line 534: if we want to reuse a existing error, would `GaugeDoesNotExist` be more appropriate?

DinizSa, line 534: I think a condition can be removed, since is already covered by the other 2 conditions
`if (end_ > _gaugesLength || start_ >= end_) ... `

DinizSa (changes requested):
By running the gas comparison report on `main` and this PR branch, the method `startDistribution` gas cost in the PR branch increased on average from `392482` to `393101` (619 gas), with 200 runs. Seems that these changes lead to a gas improvement in some conditions (ie. bigger amount of gauges?), but not in others. Imo it would be good to understand if these changes represent a saving in gas for our use case (>=11 gauges)

DinizSa (approved):
I ran some tests and with 4 plus gauges the PR changes reduce the distribution gas costs, so I guess for our use case (>11 gauges) this changes makes sense 👍

src/builderRegistry/BuilderRegistryRootstockCollective.sol
franciscotobar: I included a new error.

src/builderRegistry/BuilderRegistryRootstockCollective.sol
```diff
@@ -525,6 +526,21 @@ contract BuilderRegistryRootstockCollective is UpgradeableRootstockCollective {
         return _gauges.at(index_);
     }
 
+    /**
+     * @notice Get gauges in a specified range.
+     * @param start_ The starting index (inclusive).
+     * @param end_ The ending index (exclusive).
```
antomor, line 532: NIT: I found a bit strange the need to specify the ending index. I think it may be better to ask the users to specify how many gauges they want (
```suggestion
     * @param length_ The number of gauge addresses we want to retrieve.
```

antomor (approved):
LGTM

PR #215, opened 2025-04-23
TOK-698/halt-resume-gauges

franciscotobar:
## What

- Reduce the amount of external calls in the resume and halt gauges methods

## Why

- Going back and forth make the code complex
- Reduce gas cost

## Gas comparison
- Old
Screenshot 2025-04-23 at 10 23 14 AM

- New
Screenshot 2025-04-23 at 10 20 13 AM

## Refs

- [TOK-698](https://rsklabs.atlassian.net/browse/TOK-698)

### DIFF
--- a/src/backersManager/BackersManagerRootstockCollective.sol
+++ b/src/backersManager/BackersManagerRootstockCollective.sol
@@ -603,9 +603,15 @@ contract BackersManagerRootstockCollective is
      * produce a miscalculation of rewards
      * @param gauge_ gauge contract to be halted
      */
-    function haltGaugeShares(GaugeRootstockCollective gauge_) external onlyBuilderRegistry notInDistributionPeriod {
+    function haltGaugeShares(GaugeRootstockCollective gauge_)
+        external
+        onlyBuilderRegistry
+        notInDistributionPeriod
+        returns (uint256)
+    {
         // allocations are not considered for the reward's distribution
         totalPotentialReward -= gauge_.rewardShares();
+        return _periodFinish;
     }
 
     /**
--- a/src/backersManager/BackersManagerRootstockCollective.sol
+++ b/src/backersManager/BackersManagerRootstockCollective.sol
@@ -606,7 +606,6 @@ contract BackersManagerRootstockCollective is
     function haltGaugeShares(GaugeRootstockCollective gauge_) external onlyBuilderRegistry notInDistributionPeriod {
         // allocations are not considered for the reward's distribution
         totalPotentialReward -= gauge_.rewardShares();
-        builderRegistry.setHaltedGaugeLastPeriodFinish(gauge_, _periodFinish);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/builderRegistry/BuilderRegistryRootstockCollective.sol
```diff
@@ -677,6 +670,10 @@ contract BuilderRegistryRootstockCollective is UpgradeableRootstockCollective {
         emit CommunityApproved(builder_);
     }
 
+    function _setHaltedGaugeLastPeriodFinish(GaugeRootstockCollective gauge_, uint256 periodFinish_) internal {
```
DinizSa, line 673: [nit] I personally don't see the value in internal functions to just change a property. They increase the verbosity of the code, making it harder to read, without encapsulating any considerable logic.

src/builderRegistry/BuilderRegistryRootstockCollective.sol
```diff
@@ -604,7 +594,9 @@ contract BuilderRegistryRootstockCollective is UpgradeableRootstockCollective {
         if (!isGaugeHalted(address(gauge_))) {
             _haltedGauges.add(address(gauge_));
             _gauges.remove(address(gauge_));
-            backersManager.haltGaugeShares(gauge_);
+            BackersManagerRootstockCollective _backersManager = backersManager;
+            _backersManager.haltGaugeShares(gauge_);
```
DinizSa, line 598: we end up with the same amount of external calls, and the method `communityBanBuilder` is now slightly more expensive. A possible approach would be for the `haltGaugeShares` method to return `periodFinish`. It would reduce an external call, but not sure it makes sense such value return in terms of logic.

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -587,7 +587,6 @@ contract BackersManagerRootstockCollective is
     function haltGaugeShares(GaugeRootstockCollective gauge_) external onlyBuilderRegistry notInDistributionPeriod {
```
jurajpiar, line 587: Question about current impl:
Is it not an issue that, because of the `notInDistributionPeriod` mod, the governance call to `communityBanBuilder` may fail if in distribution?

src/builderRegistry/BuilderRegistryRootstockCollective.sol
jurajpiar: +1

src/builderRegistry/BuilderRegistryRootstockCollective.sol
```diff
@@ -178,16 +178,6 @@ contract BuilderRegistryRootstockCollective is UpgradeableRootstockCollective {
...
     // -----------------------------
 
-    function setHaltedGaugeLastPeriodFinish(
-        GaugeRootstockCollective gauge_,
-        uint256 periodFinish_
-    )
-        external
-        onlyBackersManager
-    {
-        haltedGaugeLastPeriodFinish[gauge_] = periodFinish_;
```
antomor, line 188: should we also include a gap in BuilderRegistry (and consequently a migration)?

antomor (commented):
@franciscotobar Thank you for the changes. I don't see a lot of benefit in the current implementation, but let's see if we want to move forward with it or not

src/builderRegistry/BuilderRegistryRootstockCollective.sol
franciscotobar: I wanted to keep the same methods and structure to simplify the change but I agree that I could just remove it.

src/builderRegistry/BuilderRegistryRootstockCollective.sol
franciscotobar: Why? The storage structure is the same, since I'm not moving any data. As you can see, the `haltedGaugeLastPeriodFinish` is still within `BuilderRegistry` and the `totalPotentialReward` in the `BackersManager`.

antomor (commented):
Maybe I'm missing a point here, but I don't think this PR really addresses the issue described.

antomor (approved):

PR #216, opened 2025-04-23
TOK-625: claim backers reward method per token

jurajpiar:
### In `Gauge` contract:

- makes `claimBackerReward(address,address)` internal
- adds public `claimBackerRIFReward(address)`
- and public `claimBackerRBTCReward(address)`

### In `BackersManager` contract

- removes `claimBackerRewards(address,addess)`
- adds public `claimBackerRIFReward(address)`
- adds public `claimBackerRBTCReward(address)`

## Testing

- regression on claiming backers rewards

## Refs

- [TOK-625](https://rsklabs.atlassian.net/browse/TOK-625)

### DIFF
--- a/src/backersManager/BackersManagerRootstockCollective.sol
+++ b/src/backersManager/BackersManagerRootstockCollective.sol
@@ -327,17 +327,30 @@ contract BackersManagerRootstockCollective is
     /**
      * @notice claims backer rewards from a batch of gauges
      * @param gauges_ array of gauges to claim
-     * @param rewardToken_ address of the token rewarded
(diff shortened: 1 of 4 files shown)


DELME
antomor, line 1: I guess we need to delete it 😅

antomor (commented):
There is only one ongoing conversation about the possibility to keep the same method exposed before to avoid the breaking change with the UI. Apart from this, LGTM

DELME
jurajpiar: yea, I added that to test the gas PR

antomor:
After analysing the gas impact of these changes, we decided not to merge them.

PR #217, opened 2025-04-23
Tok 706/cycle distribution simplification

DinizSa:
## What

- allow any address to fund backers manager once per cycle with the default amount

- [TOK-706](https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?assignee=712020%3A42dabdd3-0b32-4060-be21-62cfd46a1751&selectedIssue=TOK-706)
- [investigation notion](https://www.notion.so/iovlabs/TOK-578-Simplify-the-fund-distribute-steps-184c132873f980d59bc4f13a97bfd8c9?showMoveTo=true&saveParent=true)

### DIFF
--- a/src/RewardDistributorRootstockCollective.sol
+++ b/src/RewardDistributorRootstockCollective.sol
@@ -16,6 +16,7 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective 
     // -----------------------------
     error NotFoundationTreasury();
     error CollectiveRewardsAddressesAlreadyInitialized();
+    error CycleAlreadyFunded();
 
     // -----------------------------
     // --------- Modifiers ---------
@@ -23,6 +24,12 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective 
     modifier onlyFoundationTreasury() {
         governanceManager.validateFoundationTreasury(msg.sender);
         _;
+    }
+
+    modifier onlyOncePerCycle() {
+        if (lastFundedCycleStart == backersManager.cycleStart(block.timestamp)) revert CycleAlreadyFunded();
+        _;
+        lastFundedCycleStart = backersManager.cycleStart(block.timestamp);
     }
 
     // -----------------------------
@@ -37,6 +44,7 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective 
     uint256 public defaultRewardTokenAmount;
     ///@notice default reward coinbase amount
     uint256 public defaultRewardCoinbaseAmount;
+    uint256 public lastFundedCycleStart;
 
     // -----------------------------
     // ------- Initializer ---------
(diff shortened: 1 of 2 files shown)


antomor (approved):

src/RewardDistributorRootstockCollective.sol
```diff
@@ -121,17 +129,17 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective
 
     /**
      * @notice sends rewards to backersManager contract with default amounts
-     * @dev reverts if is not called by foundation treasury address
+     * @dev reverts if is called more than once per cycle
      */
-    function sendRewardsWithDefaultAmount() external payable onlyFoundationTreasury {
+    function sendRewardsWithDefaultAmount() external payable onlyOncePerCycle {
```
raullaprida, line 134: Note that whoever is calling, it's not really funding the default amounts (the amounts come from the contract). 
So my question is --> Is there a reason why you want to keep this function as **payable**?
Im not requesting for you to remove it just to explain the rationale behind it, please.

I ask because I understand you don't expect the caller EOA to pay, since msg.value  is not checked and the token balance check is not required for the caller.

src/RewardDistributorRootstockCollective.sol
```diff
@@ -121,17 +129,17 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective
...
         _sendRewards(defaultRewardTokenAmount, defaultRewardCoinbaseAmount);
     }
 
     /**
      * @notice sends rewards to backersManager contract with default amounts and starts the distribution
-     * @dev reverts if is not called by foundation treasury address
+     * @dev reverts if is called more than once per cycle
      */
-    function sendRewardsAndStartDistributionWithDefaultAmount() external payable onlyFoundationTreasury {
+    function sendRewardsAndStartDistributionWithDefaultAmount() external payable onlyOncePerCycle {
```
raullaprida, line 142: Same comments I made for `sendRewardsWithDefaultAmount()`

raullaprida (changes requested):
I left a couple of questions, that's all

src/RewardDistributorRootstockCollective.sol
antomor: The reason for this function to be payable is just to give additional flexibility to allow the caller to send RBTC directly when calling this function rather than doing it in a different tx. Although it was initially meant to be called by the Foundation, we think it still makes sense to keep it. If the caller is going to be the Foundation, they'll still have the possibility, otherwise it's up to the caller to specify the value or not. Either they do it or not, it won't change the amount sent to the BackersManager.

src/RewardDistributorRootstockCollective.sol
antomor: https://github.com/RootstockCollective/collective-rewards-sc/pull/217#discussion_r2071329770

PR #223, opened 2025-04-25
feat: gha to compare gas costs

DinizSa:
## What

- Add gh to compare functions gas costs with the base branch of a PR. The gas differences (if any) are written as a sticky comment.

### DIFF
--- a/.github/workflows/foundry-gas-diff.yml
+++ b/.github/workflows/foundry-gas-diff.yml
@@ -6,13 +6,13 @@ on:
       - main
   pull_request:
 
-permissions:
-  contents: read
-  pull-requests: write
+permissions: read-all
 
 jobs:
   compare_gas_reports:
     runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
     steps:
       - name: Harden the runner (Audit all outbound calls)
         uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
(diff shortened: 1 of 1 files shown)


.github/workflows/foundry-gas-diff.yml
```diff
@@ -0,0 +1,62 @@
...
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  contents: read
+  pull-requests: write
```
antomor, line 11: I'd move the write pull requests permissions only to the job which requires it.

.github/workflows/foundry-gas-diff.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: Rubilmax/foundry-gas-diff@60e763d02526ee3299bd04278cb178d1547b134b # v3.21
+        with:
+          summaryQuantile: 0.0 # display the 100% most significant gas diffs in the summary (defaults to 0.8 => 20%)
+          sortCriteria: avg,max # sort diff rows by criteria
+          sortOrders: desc,asc # and directions
+          ignore: test-foundry/**/* # filter out gas reports from specific paths (test/ is included by default)
+        id: gas_diff
+
+      - name: Add gas diff to sticky comment
+        if: github.event_name == 'pull_request' || github.event_name == 'pull_request_target'
```
antomor, line 57: Why do we need both `pull_request` and `pull_request_target`? In my understanding, the first should include also the latter.

antomor (approved):
@DinizSa I've reviewed it and overall looks good. I've just left few doubts/comments.

.github/workflows/foundry-gas-diff.yml
DinizSa: agree, pushed

PR #230, opened 2025-04-29
Tok 697/change the incentivize with reward token

Hakob23:
## What

- We could check for the `msg.sender` balance and allowance to `gauge` contract before calling `_notifyRewardAmount` 

## Why

- To optimize gas consumption in case of failure. Gas consumption differences are attached. 
- Gas consumption in case of failure before is `100960` and after is `36309`
[gasReport.txt](https://github.com/user-attachments/files/19958617/gasReport.txt)
[gasReportBefore.txt](https://github.com/user-attachments/files/19958618/gasReportBefore.txt)

## Testing

- `forge test --mt test_incentivizeWithNotEnoughBalance`


jurajpiar (approved):
LG

PR #241, opened 2025-05-15
Tok 729/measure gas impact of adding tokens generically

Hakob23:
## What

- Introducing new storage variables to be able to have multiple reward tokens

## Why

- With this approach it will be possible to scale the range of reward tokens that backers and builders can use

## Testing

- forge test --no-match-test '(testFuzz|invariant|fork)'

### DIFF
--- a/src/RewardDistributorRootstockCollective.sol
+++ b/src/RewardDistributorRootstockCollective.sol
@@ -10,6 +10,7 @@ import { UtilsLib } from "./libraries/UtilsLib.sol";
  * @title RewardDistributorRootstockCollective
  * @notice Accumulates all the rewards to be distributed for each cycle
  */
+
 contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective {
     // -----------------------------
     // ------- Custom Errors -------
@@ -18,7 +19,7 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective 
     error CollectiveRewardsAddressesAlreadyInitialized();
     error CycleAlreadyFunded();
     error UnequalLengths();
-    
+
     // -----------------------------
     // --------- Modifiers ---------
     // -----------------------------
--- a/src/backersManager/BackersManagerRootstockCollective.sol
+++ b/src/backersManager/BackersManagerRootstockCollective.sol
@@ -494,7 +494,7 @@ contract BackersManagerRootstockCollective is
             revert RewardTokenNotValid();
         }
     }
-    
+
     /**
      * @notice internal function used to allocate votes for a gauge or a batch of gauges
      * @param gauge_ address of the gauge where the votes will be allocated
@@ -719,7 +719,6 @@ contract BackersManagerRootstockCollective is
         // incentives can stay in the gauge because lastUpdateTime > lastTimeRewardApplicable
         // Passing an empty array as an argument with the length of the addresses
         uint256[] memory _zeroAmounts = new uint256[](2);
-
 
         if (_periodFinish <= block.timestamp) revert BeforeDistribution();
         // allocations are considered again for the reward's distribution


src/RewardDistributorRootstockCollective.sol
```diff
@@ -84,47 +113,54 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective
...
      * @notice sends rewards to backersManager contract and starts the distribution to the gauges
      * @dev reverts if is not called by foundation treasury address
      *  reverts if is not in the distribution window
-     * @param amountERC20_ amount of ERC20 reward token to send
+     * @param amountsERC20_ amount of ERC20 reward token to send
      * @param amountCoinbase_ amount of Coinbase reward token to send
      */
     function sendRewardsAndStartDistribution(
-        uint256 amountERC20_,
+        uint256[] memory amountsERC20_,
```
DinizSa, line 138: this function interface might be error prone. the amounts need to be sent in the same order as the `rewardsTokens `. Also there is no check that `amountsERC20_ ` is the same length as `rewardsTokens`, as it would fail otherwise. wdyt of receiving an extra input, as the array of the tokens addresses ?

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -631,13 +717,20 @@ contract BackersManagerRootstockCollective is
     {
         // gauges cannot be resumed before the distribution,
         // incentives can stay in the gauge because lastUpdateTime > lastTimeRewardApplicable
+        uint256[] memory _zeroAmounts = new uint256[](2);
+
+        // Passing an empty array as an argument with the length of the addresses
+        for (uint256 i = 0; i < rewardsTokens.length; i++) {
+            _zeroAmounts[i] = 0;
```
DinizSa, line 724: is this necessary? by default the array elements will already be 0

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -574,13 +653,18 @@ contract BackersManagerRootstockCollective is
     {
         uint256 _rewardShares = gauge_.rewardShares();
         // [N] = [N] * [N] / [N]
-        uint256 _amountERC20 = (_rewardShares * rewardsERC20_) / totalPotentialReward_;
+        uint256[] memory _rewardsAmounts = new uint256[](rewardsTokens.length);
         // [N] = [N] * [N] / [N]
-        uint256 _amountCoinbase = (_rewardShares * rewardsCoinbase_) / totalPotentialReward_;
+        for (uint256 i = 0; i < rewardsTokens.length; i++) {
```
DinizSa, line 658: by accessing `rewardsTokens.length` each iteration to check if `i` reached the max, it performs a SLOAD on every iteration. can create memory variable with length

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -489,22 +570,26 @@ contract BackersManagerRootstockCollective is
      * @return true if distribution has finished
      */
     function _distribute() internal returns (bool) {
+        uint256[] memory _amounts = new uint256[](rewardsTokens.length);
+        for (uint256 i = 0; i < rewardsTokens.length; i++) {
```
DinizSa, line 574: same as [here](https://github.com/RootstockCollective/collective-rewards-sc/pull/241/files#r2095435435)

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -592,8 +676,10 @@ contract BackersManagerRootstockCollective is
      * @param value_ amount of rewardTokens to approve
      */
     function rewardTokenApprove(address gauge_, uint256 value_) external onlyBuilderRegistry {
-        if (!IERC20(rewardToken).approve(gauge_, value_)) {
-            revert RewardTokenNotApproved();
+        for (uint256 i = 0; i < rewardsTokens.length; i++) {
```
DinizSa, line 679: same as [here](https://github.com/RootstockCollective/collective-rewards-sc/pull/241/files#r2095435435)

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -244,7 +277,9 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
      * @param backer_ address who receives the rewards
      */
     function claimBackerReward(address backer_) external {
-        claimBackerReward(rewardToken, backer_);
+        for (uint256 i = 0; i < rewardsTokens.length; i = UtilsLib._uncheckedInc(i)) {
```
DinizSa, line 280: same as [here](https://github.com/RootstockCollective/collective-rewards-sc/pull/241/files#r2095435435)

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -277,7 +312,9 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
      * @dev rewards are transferred to the builder reward receiver
      */
     function claimBuilderReward() external {
-        claimBuilderReward(rewardToken);
+        for (uint256 i = 0; i < rewardsTokens.length; i = UtilsLib._uncheckedInc(i)) {
```
DinizSa, line 315: same as [here](https://github.com/RootstockCollective/collective-rewards-sc/pull/241/files#r2095435435)

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -335,12 +380,15 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
         // if backers quit before cycle finish we need to store the remaining rewards on first allocation
         // to add it on the next reward distribution
         if (totalAllocation == 0) {
-            _updateRewardMissing(rewardToken, _periodFinish);
             _updateRewardMissing(UtilsLib._COINBASE_ADDRESS, _periodFinish);
         }
-
-        _updateRewards(rewardToken, backer_, _periodFinish);
         _updateRewards(UtilsLib._COINBASE_ADDRESS, backer_, _periodFinish);
+        for (uint256 i = 0; i < rewardsTokens.length; i++) {
```
DinizSa, line 386: same as [here](https://github.com/RootstockCollective/collective-rewards-sc/pull/241/files#r2095435435)

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -423,15 +451,17 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
...
      * @param periodFinish_ timestamp end of current rewards period
      * @param cycleStart_ Collective Rewards cycle start timestamp
      * @param cycleDuration_ Collective Rewards cycle time duration
      * @return newGaugeRewardShares_ new gauge rewardShares, updated after the distribution
      */
+    // TODO: add comments
     function notifyRewardAmountAndUpdateShares(
-        uint256 amountERC20_,
+        uint256[] calldata amountsERC20_,
+        address[] calldata addressesERC20_,
```
DinizSa, line 464: nit: addresses in the name might not be needed, as the type already gives that information. maybe a more descriptive erc20_ or erc20Tokens_? if the second is chosen, maybe rename also ERC20Amounts_ to keep consistency

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -89,6 +90,19 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
     /// @notice BuilderRegistryRootstockCollective contract address
     BuilderRegistryRootstockCollective public builderRegistry;
 
+    // -----------------------------
+    // -------- V3 Storage ---------
+    // -----------------------------
+
+    /// @notice addresses of all valid rewards tokens
+    address[] public rewardsTokens;
```
DinizSa, line 98: this is in double plural, `rewardTokens` sound better

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -89,6 +90,19 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
...
+
+    /// @notice mapping of validated reward tokens
+    mapping(address => bool) public rewardsTokensValid;
+
+    /// @notice mapping of reward amounts of reward tokens
+    mapping(address => uint256) public rewardsAmounts;
```
DinizSa, line 104: this is not being used in this file, is it needed?

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -93,6 +94,11 @@ contract BackersManagerRootstockCollective is
         _;
     }
 
+    modifier MoreThanZeroGauges() {
```
DinizSa, line 97: maybe an alternative shorter name `hasGauges`
also usually modifiers are camel case.

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -532,7 +611,11 @@ contract BackersManagerRootstockCollective is
         if (_lastDistribution == _gaugesLength) {
             _finishDistribution();
             totalPotentialReward = _newTotalPotentialReward;
-            rewardsERC20 = rewardsCoinbase = 0;
+            for (uint256 i = 0; i < rewardsERC20_.length; i++) {
+                // Storage rewards are getting updated
+                rewardsAmounts[rewardsTokens[i]] = 0;
```
DinizSa, line 616: why are `rewardsERC20_` passed as parameter? just the length of the array is used, to iterate over the rewardsAmounts...

src/gauge/GaugeRootstockCollective.sol
antomor: Should we specify that this are all the ERC20 tokens only? or do we want to keep track of RBTC also here?

antomor: I'd keep RBTC and erc20 split, so we don't need to check them each time, unless, we consider that the first position is the RBTC and the others are ERC20

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -301,14 +338,22 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
         }
     }
 
+    function validateRewardToken(address rewardToken_) public view {
```
antomor, line 341: Why do we need this function to be public? couldn't it be private or internal?

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -442,33 +472,36 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
...
-            _rewardToken,
-            amountERC20_ - _backerAmountERC20,
-            _backerAmountERC20,
-            periodFinish_,
-            _timeUntilNextCycle,
-            _resetRewardMissing
-        );
+        // TODO: Check length and revert if not equal
+        uint256 _timeUntilNextCycle = UtilsLib._calcTimeUntilNextCycle(cycleStart_, cycleDuration_, block.timestamp);
+        for (uint256 i = 0; i < amountsERC20_.length; i++) {
```
antomor, line 478: This could be optimizes with local variable and unchecked math

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -442,33 +472,36 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
...
             _timeUntilNextCycle,
-            _resetRewardMissing
+            true /*_resetRewardMissing*/
         );
 
         newGaugeRewardShares_ = totalAllocation * cycleDuration_;
         rewardShares = newGaugeRewardShares_;
 
-        SafeERC20.safeTransferFrom(IERC20(_rewardToken), msg.sender, address(this), amountERC20_);
+        for (uint256 i = 0; i < addressesERC20_.length; i++) {
```
antomor, line 502: This can be optimized

src/gauge/GaugeRootstockCollective.sol
Hakob23: I have made it public as an extra feature for the users to validate the token before trying to incentivize with it, as this is the more generic version. But if this feature is not required now, it can be made internal.

src/backersManager/BackersManagerRootstockCollective.sol
Hakob23: Yes, you are right. The time of writing this function, the Gauge contract wasn't including the new storage variables, and after introducing them, I didn't want to change the structure of the code not to mess up with the signatures.

src/RewardDistributorRootstockCollective.sol
Hakob23: I agree with implementing sameLength check in rewardDistributor as well(although it's another duplicate piece of code). I was thinking to pass the addresses as an input as well, but left the current signatures of functions unchanged just to show the PoC code

src/gauge/GaugeRootstockCollective.sol
Hakob23: For `notifyRewardAmountAndUpdateShares` function, we are hitting `stackTooDeep` error, if adding even a single additional variable

src/gauge/GaugeRootstockCollective.sol
Hakob23: making internal for now

src/gauge/GaugeRootstockCollective.sol
Hakob23: `rewardsAmounts` is not used, and has been removed. `rewardsTokensValid` is used to validate the token when incentivizing with it.

src/gauge/GaugeRootstockCollective.sol
Hakob23: Yes, we do assume that the first token is RBTC as it is being passed as the first element during the initialization proccesses.

src/gauge/GaugeRootstockCollective.sol
Hakob23: changed it to tokensERC20_

src/RewardDistributorRootstockCollective.sol
Hakob23: I have changed my mind on passing the addresses as an argument. We can check them by querying the array, then passing the amounts in corresponding order.

src/RewardDistributorRootstockCollective.sol
```diff
@@ -42,11 +44,24 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective
     /// @notice BackersManagerRootstockCollective contract address
     BackersManagerRootstockCollective public backersManager;
     ///@notice default reward token amount
-    uint256 public defaultRewardTokenAmount;
+    uint256[] public defaultRewardTokenAmount;
```
raullaprida, line 47: The name "default..." and now being a list, doesn't really apply. The variable should have a new name ideally

src/RewardDistributorRootstockCollective.sol
```diff
@@ -84,47 +115,54 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective
...
      * @notice sends rewards to backersManager contract to be distributed to the gauges
      * @dev reverts if is not called by foundation treasury address
-     * @param amountERC20_ amount of ERC20 reward token to send
+     * @param amountsERC20_ amount of ERC20 reward token to send
      * @param amountCoinbase_ amount of Coinbase reward token to send
      */
-    function sendRewards(uint256 amountERC20_, uint256 amountCoinbase_) external payable onlyFoundationTreasury {
-        _sendRewards(amountERC20_, amountCoinbase_);
+    function sendRewards(
+        uint256[] memory amountsERC20_,
```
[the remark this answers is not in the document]
raullaprida: At the minimum we need to let it clear in the function's API/function description. Nevertheless, it would be quite error prone and probably "sort of low level" for the average user to have to match the order of the values to the internal order of the token list

src/backersManager/BackersManagerRootstockCollective.sol
```diff
@@ -414,6 +488,13 @@ contract BackersManagerRootstockCollective is
     // ---- Internal Functions -----
     // -----------------------------
 
+    // TODO: Add comments
+    function _validateRewardToken(address rewardToken_) internal view {
```
raullaprida, line 492: you're not validating it, you're verifying it is valid

src/RewardDistributorRootstockCollective.sol
```diff
@@ -42,11 +44,24 @@ contract RewardDistributorRootstockCollective is UpgradeableRootstockCollective
...
+    // -----------------------------
+
+    /// @notice addresses of all valid rewards tokens
+    address[] public rewardTokens;
+
+    /// @notice mapping of validated reward tokens
+    mapping(address => bool) public rewardTokensValid;
+
+    /// @notice mapping of reward amounts of reward tokens
+    mapping(address => uint256) public rewardsAmounts;
```
raullaprida, line 63: Is `rewardsAmounts` being used? if so, where.

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -277,7 +310,10 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
      * @dev rewards are transferred to the builder reward receiver
      */
     function claimBuilderReward() external {
-        claimBuilderReward(rewardToken);
+        uint256 _rewardTokensLength = rewardTokens.length;
+        for (uint256 i = 0; i < _rewardTokensLength; i = UtilsLib._uncheckedInc(i)) {
```
raullaprida, line 314: Same comment as in claimBackerRewards. There's no option to specify which tokens to claim. Difficult scenario if the token list is huge and the builder has rewards in only one or two

PR #272, opened 2025-09-02
Tok 942/upgrade script usdrif

DinizSa:
## What

- Update the upgrade script after the USDRIF changes

## Why

- so we can properly upgrade the live smart contracts

## Testing

- `bun run test:fork`

## Refs

- [Ticket](https://rsklabs.atlassian.net/jira/software/projects/TOK/boards/267?selectedIssue=TOK-942)


test/upgradeV3/UpgradeV3.t.sol
```diff
@@ -154,15 +171,160 @@ contract UpgradeV3Test is Test {
...
 
         // THEN rewardDistributor should have the new implementation
-        vm.assertEq(_getImplementation(address(rewardDistributor)), address(upgradeV3.rewardDistributorImplV3()));
+        vm.assertEq(_getImplementation(address(rewardDistributorV3)), address(upgradeV3.rewardDistributorImplV3()));
+        // AND usdrifToken should be initialized with the same token as backersManagerV2
+        vm.assertEq(address(rewardDistributorV3.usdrifToken()), address(backersManagerV3.usdrifToken()));
+        vm.assertEq(address(rewardDistributorV3.usdrifToken()), usdrifToken);
+        // AND rifToken should be initialized with the same token as backersManagerV3
+        vm.assertEq(address(rewardDistributorV3.rifToken()), address(backersManagerV3.rifToken()));
+        vm.assertNotEq(address(rewardDistributorV3.rifToken()), address(0));
```
antomor, line 188: why don't we compare it here against the real RIF token rather than checking it's not zero?

test/upgradeV3/UpgradeV3.t.sol
```diff
@@ -154,15 +171,160 @@ contract UpgradeV3Test is Test {
...
+        // GIVEN the rewardDistributor is still v2
+        // THEN it should revert when accessing usdrifToken
+        vm.expectRevert();
+        RewardDistributorRootstockCollective(payable(address(rewardDistributorV2))).usdrifToken();
+    }
+
+    /**
+     * SCENARIO: rewardDistributor initializeV3 sets usdrifToken correctly during upgrade
+     */
+    function test_fork_rewardDistributorinitializeV3() public {
```
antomor, line 223:
```suggestion
    function test_fork_rewardDistributorInitializeV3() public {
```

test/upgradeV3/UpgradeV3.t.sol
```diff
@@ -154,15 +171,160 @@ contract UpgradeV3Test is Test {
...
+        // GIVEN the contracts are not yet upgraded
+        // WHEN the upgrade is performed
+        _upgradeV3();
+
+        // THEN usdrifToken should be properly initialized
+        vm.assertNotEq(address(rewardDistributorV3.usdrifToken()), address(0));
```
antomor, line 229: I'd remove this since we check it against the token address just in the line after

test/upgradeV3/UpgradeV3.t.sol
```diff
@@ -154,15 +171,160 @@ contract UpgradeV3Test is Test {
...
+        vm.assertEq(_gaugeFactory.usdrifToken(), usdrifToken);
+
+        // AND GaugeBeacon should be upgraded to the new implementation
+        vm.assertEq(gaugeBeacon.implementation(), address(upgradeV3.gaugeImplV3()));
+
+        // AND Gauge V3 functions should be accessible (not revert)
+        GaugeRootstockCollective _gaugeV3 = GaugeRootstockCollective(address(gaugeBeacon.implementation()));
+        // return data is zero as it is just the proxy implementation, without storage set
+        _gaugeV3.rifToken();
+        _gaugeV3.usdrifToken();
```
antomor, line 268: should we add 2 assertions here?

test/upgradeV3/UpgradeV3.t.sol
```diff
@@ -154,15 +171,160 @@ contract UpgradeV3Test is Test {
...
+        uint256 _gaugesLength = builderRegistry.getGaugesLength();
+
+        // THEN each gauge should have rifToken and usdrifToken properly initialized
+        for (uint256 i = 0; i < _gaugesLength; i++) {
+            address _gaugeAddress = builderRegistry.getGaugeAt(i);
+            GaugeRootstockCollective _gauge = GaugeRootstockCollective(_gaugeAddress);
+
+            // Check that rifToken exists and is not zero address
+            address _rifToken = _gauge.rifToken();
+            vm.assertNotEq(_rifToken, address(0), "rifToken should not be zero address");
```
antomor, line 288: same as before, should we check it's exactly a specific address?

test/upgradeV3/UpgradeV3.t.sol
DinizSa: I think that we we are doing in the 1867. line 188 was more to check that in test environments we revert if both addresses (`rewardDistributorV3.rifToken` and `backersManagerV3.rifToken`) are 0.
However, I will just ensure that `backersManagerV3.rifToken` is not zero on the specific  backers manager upgrade test.

test/upgradeV3/UpgradeV3.t.sol
DinizSa: since we are calling the implementation just to check that it has the new methods, it will always retrieve 0. but I'll improve the test like so:
```
// AND rifToken and usdrifToken should be zero address as _gaugeV3 is the implementation, not the proxy, so there is no storage set
        vm.assertEq(_gaugeV3.rifToken(), address(0));
        vm.assertEq(_gaugeV3.usdrifToken(), address(0));
```

antomor (approved):

PR #282, opened 2025-09-29
refactor: emit events with cycle rewards distribution amounts

franciscotobar:
## What

- Emit the amount of rewards distributed during each cycle in the `BackersManager`
- Emit when the default amounts are updated in the `RewardDistributor`

## Why

- Information that is necessary for the subgprahs

## Testing

- `bun run test`

## Refs

- [TOK-967](https://rsklabs.atlassian.net/browse/TOK-967)


test/RewardDistributorRootstockCollective.t.sol
```diff
@@ -80,6 +85,27 @@ contract RewardDistributorRootstockCollectiveTest is BaseTest {
...
+     * SCENARIO: set default reward amounts
+     */
+    function test_SetDefaultRewardAmount() public {
+        // GIVEN a RewardDistributorRootstockCollective contract
+        vm.startPrank(foundation);
+
+        // WHEN default reward amounts are set
+        //  THEN DefaultRewardAmountsUpdated event is emitted
+        vm.expectEmit();
+        emit DefaultRewardAmountsUpdated(10 ether, 1 ether, 10 ether);
```
DinizSa, line 98: nit: can create variables for each value (ie. `uint256 usdRifDefault = 10`, just to make the test more readable and less error prone

PR #314, opened 2026-06-29
Chore/gauge initialize

DinizSa:
## What

- Updated BackersManager and GaugeRootstockCollective implementations with initialization improvements


deployments/mainnet/BackersManagerRootstockCollective.abi.json
DinizSa, line 1: although there are abi changes, this files were not commonly updated in the past in this project. updating them now for mainnet.

deployments/mainnet/GaugeRootstockCollectiveImplementation.abi.json
DinizSa, line 1: this file was not generated in the past in this project. updating them now for mainnet.

src/upgrades/UpgradeV3.sol
```diff
@@ -110,19 +110,7 @@ contract UpgradeV3 {
...
-        }
-
-        // Initialize halted gauges with usdrifToken
-        uint256 _haltedGaugesLength = builderRegistryProxy.getHaltedGaugesLength();
-        for (uint256 i = 0; i < _haltedGaugesLength; i++) {
-            address _gaugeAddress = builderRegistryProxy.getHaltedGaugeAt(i);
-            GaugeRootstockCollective(_gaugeAddress).initializeV3(usdrifToken);
-        }
+        // Gauges already have rifToken and usdrifToken set during `initialize()`.
     }
```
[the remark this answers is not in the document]
DinizSa: upgrade v3 logic will be removed. we choose to not remove it here, to keep this process as simple as possible

src/gauge/GaugeRootstockCollective.sol
```diff
@@ -124,18 +124,9 @@ contract GaugeRootstockCollective is ReentrancyGuardUpgradeable {
...
-     * @notice For more info on supported tokens, see:
-     * https://github.com/RootstockCollective/collective-rewards-sc/blob/main/README.md#Reward-token
-     */
-    function initializeV3(address usdrifToken_) external reinitializer(3) {
-        __ReentrancyGuard_init();
-        usdrifToken = usdrifToken_;
-    }
+    // NOTE: `initializeV3()` was removed after the V3 migration. Gauges that were migrated during UpgradeV3
+    // already reached version 3. New gauges remain at version 1 after `initialize()`; this is safe because
+    // no public reinitializer remains callable.
```
aurersk: But old ones will have a different version, so if we need to update both old and new aftwerwards it will be a mess. We should make the old and new instances indistinguishable if we look at the values of their instance variables. That means both new and old need to be at the same initialize version. And given the the version number should only increase, all versions should be 3 at least.

aurersk: Commit d5a83992675f81c94d594baae7554428990a7fb2 makes this inconsistence go away.

franciscotobar (approved):
LGTM

